PR #1, opened 2025-01-17
Semantic Variables 

i-subham23:


frontend-sdk/packages/react-ui/src/components/Input/Input.tsx
```diff
@@ -0,0 +1,23 @@
...
+  className?: string;
+  size?: "small" | "medium" | "large";
+}
+
+export const Input = React.forwardRef<HTMLInputElement, InputProps>(
+  ({ className, styles, size = "medium", ...props }, ref) => {
+    return (
+      <input
+        ref={ref}
+        className={clsx("crayon-input", `crayon-input-${size}`, className)}
```
thesysHrijul, line 15: size can be undefined. this would create an invalid class name

PR #2, opened 2025-01-22
Components by AD

ankit-thesys:


frontend-sdk/packages/react-ui/src/components/ListItem/ListItem.tsx
```diff
@@ -0,0 +1,38 @@
+import clsx from "clsx";
+import { type LucideIcon } from "lucide-react";
+import React from "react";
+
+export interface ListItemProps {
+  className?: string;
+  style?: React.CSSProperties;
+  decorativeIcon?: LucideIcon;
```
thesysHrijul: Let's take `React.ReactNode` instead of these everywhere. We can let the user have the flexibility to use their own icons as well. We're also already taking icons as react nodes in a few other crayon components so we should be consistent across all components.

PR #7, opened 2025-01-27
Accordion component

i-subham23:


frontend-sdk/packages/react-ui/src/components/Accordion/accordion.scss
```diff
@@ -0,0 +1,156 @@
+@use "sass:map";
+@use "../../cssUtils" as cssUtils;
+
+/* reset */
+button,
+h3 {
+  all: unset;
+}
+
```
ankit-thesys: do not reset on the element level reset on the component class

PR #41, opened 2025-02-18
Bump version to 0.6.0

Aditya-thesys:


rabisg:
there is a new package stream now. just fyi

PR #71, opened 2025-03-06
Stream improvements

rabisg:
- Fix bug in serialization
- Add helpers to work with templates

### DIFF
--- a/js/packages/stream/src/types.ts
+++ b/js/packages/stream/src/types.ts
@@ -1,7 +1,3 @@
-const escapeString = (s: string) => {
-  return s.replaceAll('"', '\\"').replaceAll("\n", "\\n");
-};
-
 interface Chunk {
   toSSEString(): string;
 }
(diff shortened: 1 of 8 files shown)


js/packages/stream/src/types.ts
```diff
@@ -23,6 +19,6 @@ export class ResponseTemplateChunk implements Chunk {
   constructor(private readonly template: ResponseTemplate) {}
 
   toSSEString(): string {
-    return `1:${escapeString(JSON.stringify(this.template))}\n`;
+    return `1:${JSON.stringify(this.template)}\n`;
```
abhithesys, line 22: why remove escapeString?

abhithesys (approved):

js/packages/stream/src/types.ts
rabisg: we were not using it anywhere - we will probably need to bring it back when we move to SSE

PR #83, opened 2025-03-14
Update SSE format in stream pkg

abhithesys:


js/packages/react-core/src/utils/processStreamedMessage.ts
abhithesys, line 1: moved this file stream.
1. makes sense to keep it there
2. resulted in circular dependency in package.json, won't be able import SSEType enum

js/packages/stream/src/processStreamedMessage.ts
```diff
@@ -0,0 +1,115 @@
+import { AssistantMessage } from "@crayonai/react-core/src/types";
+import { createParser } from "eventsource-parser";
```
abhithesys: as discussed, used an external pkg for parsing and encoding

js/packages/stream/src/transformer.ts
```diff
@@ -14,38 +14,39 @@ export class CrayonDataStreamTransformer implements TransformStream<InputType, O
 
   constructor(opts?: TransformerOpts) {
```
abhithesys, line 16: simplified the logic

js/packages/stream/src/types.ts
```diff
@@ -1,3 +1,13 @@
+import { encode } from "eventsource-encoder";
```
abhithesys, line 1: updated to use pkg for encoding and added support for ResponseTemplatePropsStream

js/packages/stream/src/types.ts
```diff
@@ -1,3 +1,13 @@
...
+
+export enum SSEType {
+  TextDelta = "T",
```
rabisg: lets keep it a little more descriptive
- "text"/"text_delta"
- "tpl_name"
- "tpl_args"/"tpl_args"delta" (lets use args instead of props to keep it a little more generic)
- "context"/"context_delta"
- "message_update" (we can keep it message update and pass a partial message which may or may not include id)

js/packages/stream/src/types.ts
```diff
@@ -19,6 +32,18 @@ export class ResponseTemplateChunk implements Chunk {
   constructor(private readonly template: ResponseTemplate) {}
 
   toSSEString(): string {
-    return `1:${JSON.stringify(this.template)}\n`;
+    return encode({
```
rabisg, line 64: @abhithesys lets create a class for each of these SSETypes?

js/packages/stream/src/processStreamedMessage.ts
```diff
@@ -0,0 +1,115 @@
+import { AssistantMessage } from "@crayonai/react-core/src/types";
```
rabisg: this is a little problematic.
Depending on react-core would mean depending on react which would mean that backend repos will now start depending on react

js/packages/stream/src/processStreamedMessage.ts
```diff
@@ -0,0 +1,115 @@
...
+
+    if (!isMessagePushed) {
+      createMessage(message);
+    } else {
+      updateMessage(message);
+    }
+    isMessagePushed = true;
+  };
+
+  let messageContent: NonNullable<AssistantMessage["message"]> = [];
```
rabisg: what the benefit of writing NonNullable? it won't really get checked at runtime

rabisg (commented):
Can we do this structure:
1. pkg/stream - has minimal dependencies
2. pkg/react-core - depends on pkg/stream
3. pkg/react-ui - depends on both pkg/stream an pkg/react-core

We need to support Python too so we should keep pkg/stream such that we can write an equivalent package in py and other languages

js/packages/stream/src/processStreamedMessage.ts
abhithesys: I m always assigning some value to the variable but the type declaration includes undefined. NonNullable removes the undefined part from the type.

js/packages/react-core/src/stream/processStreamedMessage.ts
```diff
@@ -0,0 +1,133 @@
...
+          break;
+        }
+        case SSEType.ResponseTemplateArgs: {
+          const responseTemplateArgs = ResponseTemplateArgsChunk.fromSSEData(event.data);
+          invariant(lastMessageContent?.type === "template", "response template expected");
+          messageContent.pop();
+          messageContent = messageContent.concat({
+            type: "template",
+            name: lastMessageContent.name,
+            // XXX: This doesn't look right.
```
rabisg: @abhithesys is this function right? where do we parse the templateProps.content partial string into something that can be passed to a react component?

abhithesys: we don't parse the `templateProps.content`.
it is upto the user how they want to process it.

in case of C1, the C1 component has the responsibility of parsing the content.

rabisg (approved):

PR #87, opened 2025-03-18
fix(textSchema): fix text response schema

thesysHrijul:


js/packages/stream/src/textSchema.ts
```diff
@@ -4,8 +4,7 @@
 export const TextResponseSchema = {
   type: "object",
   properties: {
-    type: { const: "text" },
-    description: "Use this as a fallback when no other response template is applicable",
+    type: { const: "text", description: "Use this as a fallback when no other response template is applicable" },
```
rabisg, line 7: this should be at the top level for this entire schema

thesysHrijul: @rabisg Got it. But discussed with @abhithesys and closing this PR since you've already made this change in yours.

PR #119, opened 2025-04-09
Message context guide

thesysHrijul:


thesysHrijul:
TODO:

- [ ] Remove the usage of the term 'Message Context' from the entire guide and rephrase to refer to `useMessage` instead.
- [ ] Remove the schema codeblock with an inline `z.object({})`.

abhithesys (approved):
well explained

PR #121, opened 2025-04-10
Update the streaming protocol / data format documentation

thesysHrijul:
- Updated the documentation for the streaming protocol and data format


abhithesys (approved):

abhithesys:
lets add reference to [event stream helpers](https://github.com/thesysdev/crayon/blob/main/js/packages/stream/src/types.ts)

PR #138, opened 2025-06-17
Support React 17

rabisg:


thesysHrijul:
## Primary Changes

- Use a `use-sync-external-store` polyfill package provided by react
- Add a `useId` polyfill
- Downgrade `zustand` version to one that does not require react 18 as the minimum react version..

thesysHrijul (approved):

PR #168, opened 2025-07-27
Adding new bg-color variable

i-subham23:
1. Adding a new bg-colour variable
2. Adding a new sm-size variant to the tag component

### DIFF
--- a/js/packages/react-ui/package.json
+++ b/js/packages/react-ui/package.json
@@ -2,7 +2,7 @@
   "type": "module",
   "name": "@crayonai/react-ui",
   "license": "MIT",
-  "version": "0.8.16",
+  "version": "0.8.17",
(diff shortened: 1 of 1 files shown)


js/packages/react-ui/src/components/ThemeProvider/themePresets.ts
```diff
@@ -46,6 +46,7 @@ export const themePresets = {
       invertedFills: "rgba(21, 21, 21, 1)",
       elevatedFills: "rgba(255, 255, 255, 0.6)",
       sunkBgFills: "rgba(0, 0, 0, 0.04)",
+      highlightSubtle: "rgba(0, 0, 0, 0.03)",
```
abhithesys: I don't think we need to add this here

PR #189, opened 2025-09-01
Update Carousel component styles for improved layout and responsiveness

i-subham23:


js/packages/react-ui/src/components/Carousel/carousel.scss
```diff
@@ -61,4 +61,10 @@
       right: -12px;
     }
   }
+  @media (max-width: 400px) {
```
abhithesys, line 64: is this for mobile?

PR #244, opened 2025-12-21
Add PrintContext

abhithesys:
For printing a component several changes might be required like changing layout, hide something, reveal hidden elements.

This change adds PrintContext, which devs can use to tell components that they are being printed.
All the chart animations have been disabled, when in print context.

### DIFF
--- a/js/packages/react-ui/src/components/Charts/PieChart/PieChart.tsx
+++ b/js/packages/react-ui/src/components/Charts/PieChart/PieChart.tsx
@@ -1,5 +1,5 @@
 import clsx from "clsx";
-import { memo, useCallback, useEffect, useLayoutEffect, useMemo, useRef, useState } from "react";
+import { memo, useCallback, useLayoutEffect, useMemo, useRef, useState } from "react";
 import { Cell, Pie, PieChart as RechartsPieChart } from "recharts";
 import { usePrintContext } from "../../../context/PrintContext.js";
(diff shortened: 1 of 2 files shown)


js/packages/react-ui/src/components/Charts/PieChart/PieChart.tsx
```diff
@@ -285,6 +289,16 @@ const PieChartComponent = <T extends PieChartData>({
...
+  // Read initial dimensions synchronously before first paint to avoid size jump
+  useLayoutEffect(() => {
+    const wrapper = wrapperRef.current;
+    if (!wrapper) return;
+
+    const rect = wrapper.getBoundingClientRect();
+    setWrapperRect({ width: rect.width, height: rect.height });
+  }, []);
+
+  // Use ResizeObserver for subsequent size changes
```
ankit-thesys: why are we doing this ?

ankit-thesys: and if so why not do the same for radial chart ?

abhithesys: I am using useLayoutEffect to set the wrapper size. This reduces the flickering of the chart.

1. Moved the logic to the place we are using resize observer.
2. Added the same to radar chart

PR #334, opened 2026-03-14
React Native example

abhithesys:
<video
    src="https://github.com/user-attachments/assets/cff424eb-c82b-443f-bd77-2dc8bae8215e"
    noControls
    playsInline
    muted
    preload="metadata"
    style="height:600px"
    autoPlay
    loop
  />

### DIFF
--- a/examples/openui-react-native/backend/src/app/api/chat/route.ts
+++ b/examples/openui-react-native/backend/src/app/api/chat/route.ts
@@ -3,7 +3,7 @@ import { NextRequest } from "next/server";
 import OpenAI from "openai";
 import { join } from "path";
 
-const openai = new OpenAI({ apiKey: process.env.OPENAI_API_KEY });
+const openai = new OpenAI({ apiKey: process.env.OPENAI_API_KEY ?? "" });
 
 const SYSTEM_PROMPT = readFileSync(join(process.cwd(), "src", "system-prompt.txt"), "utf-8");
 


docs/content/docs/openui-lang/examples/react-native.mdx
```diff
@@ -0,0 +1,196 @@
+---
+title: React Native
+description: Stream OpenUI Lang to a React Native (Expo) app using a Next.js backend.
+---
+
+A full-stack example that streams [OpenUI Lang](/docs/openui-lang/overview) into an Expo app using a Next.js backend, then renders it as native React Native UI with `<Renderer />` from `@openuidev/react-lang`.
+
+[View source on GitHub →](https://github.com/thesysdev/crayon/tree/main/examples/openui-react-native)
```
rabisg, line 8: link needs updating

docs/content/docs/openui-lang/examples/react-native.mdx
```diff
@@ -0,0 +1,196 @@
...
+    PieChartComponent,
+    CardComponent,
+  ],
+  root: "Card",
+});
+```
+
+The mobile version contains the real React Native renderers. It is the code that actually turns parsed OpenUI Lang nodes into UI on screen.
+
+See [Defining Components](/docs/openui-lang/defining-components) for the full `defineComponent` API and [The Renderer](/docs/openui-lang/renderer) for how the library is used at runtime.
```
rabisg, line 137: urls need to updated

docs/content/docs/openui-lang/examples/react-native.mdx
```diff
@@ -0,0 +1,196 @@
...
+
+Keeping streamed content in a separate store means each bubble can update independently instead of forcing the entire message list to re-render on every chunk.
+
+See [The Renderer](/docs/openui-lang/renderer) for the full `<Renderer />` API.
+
+### `backend/src/library.ts` - Node-compatible prompt source
+
+This file mirrors the mobile library, but every `component` returns `null`. That lets the OpenUI CLI inspect the component definitions in a Node environment and generate `system-prompt.txt` without importing React Native.
+
+Run `pnpm generate:prompt` any time you change:
```
rabisg, line 174: given that we are already duplicating the types - do we need this?
cant we directly call library.toPrompt() on the backend in nextjs

docs/content/docs/openui-lang/examples/react-native.mdx
```diff
@@ -0,0 +1,196 @@
...
+<Renderer response={openui} library={library} isStreaming={isStreaming} />
+```
+
+Keeping streamed content in a separate store means each bubble can update independently instead of forcing the entire message list to re-render on every chunk.
+
+See [The Renderer](/docs/openui-lang/renderer) for the full `<Renderer />` API.
+
+### `backend/src/library.ts` - Node-compatible prompt source
+
+This file mirrors the mobile library, but every `component` returns `null`. That lets the OpenUI CLI inspect the component definitions in a Node environment and generate `system-prompt.txt` without importing React Native.
```
rabisg, line 172: can we have a common file that both backend and react-native imports that just has the zod types rather than duplicating it

examples/openui-react-native/backend/src/library.ts
```diff
@@ -0,0 +1,123 @@
...
+    TextComponent,
+    BarChartComponent,
+    LineChartComponent,
+    PieChartComponent,
+    CardComponent,
+  ],
+  root: "Card",
+});
+
+export const promptOptions: PromptOptions = {
```
rabisg, line 107: do we need this? Doesn't the toPrompt already handle this

examples/openui-react-native/chat-app/components/BotBubble.tsx
```diff
@@ -0,0 +1,37 @@
+import { Renderer } from "@openuidev/react-lang";
+import React from "react";
+import { ActivityIndicator, StyleSheet, View } from "react-native";
+import { library } from "../library";
+import { useStreamContent } from "../store/streamStore";
+
+export function BotBubble({ messageId }: { messageId: string }) {
```
rabisg, line 7: would it be better to call this AssistantBubble & AssistantMessage throughout rather than calling it Bot to keep it consistent with other examples

docs/content/docs/openui-lang/examples/react-native.mdx
abhithesys: facing react-version mismatch issue in next. tried building it outside of pnpm-package, issue remained. we either ways need to fix the approach

examples/openui-react-native/chat-app/screens/ChatScreen.tsx
```diff
@@ -0,0 +1,101 @@
...
+import { ChatInput } from "../components/ChatInput";
+import { SuggestedPrompts } from "../components/SuggestedPrompts";
+import { UserBubble } from "../components/UserBubble";
+import { StreamMessage, useStreamingChat } from "../hooks/useStreamingChat";
+import { pushStream } from "../store/streamStore";
+
+// ─── Message types ────────────────────────────────────────────────────────────
+
+type UserMessage = { id: string; type: "user"; text: string };
+type BotMessage = { id: string; type: "bot" };
```
rabisg, line 14: same comment: Assistant rather than Bot

docs/content/docs/openui-lang/examples/react-native.mdx
abhithesys: cli isn't working with react-native components and importing in next is also broken at the moment.

examples/openui-react-native/chat-app/app.json
```diff
@@ -0,0 +1,30 @@
...
+      "resizeMode": "contain",
+      "backgroundColor": "#ffffff"
+    },
+    "ios": {
+      "supportsTablet": true
+    },
+    "android": {
+      "adaptiveIcon": {
+        "backgroundColor": "#E6F4FE",
+        "foregroundImage": "./assets/android-icon-foreground.png",
```
rabisg, line 20: can we add the iOS images and/or remove the android ones

docs/content/docs/openui-lang/examples/react-native.mdx
rabisg: my bad. I thought Im reading the README. these links are fine

abhithesys: the urls are correct:
https://www.openui.com/docs/openui-lang/renderer
https://www.openui.com/docs/openui-lang/defining-components

PR #338, opened 2026-03-14
Fix/react headless OpenAI sse decoder

boffin-dmytro:
## What’s in this PR
This PR makes OpenAI streaming in `@openuidev/react-headless` more reliable by introducing a small internal SSE decoder and using it in the OpenAI SSE adapters.

Previously, the OpenAI adapters parsed streams by splitting each decoded network chunk on `\n` and treating each `data:` line as complete JSON. In real-world streaming, chunk boundaries are arbitrary (JSON can be split mid-token), and SSE events are framed by blank lines and can include multi-line `data:` fields.

## Changes
- Added an internal SSE decoder (`decodeSSE`) that:
  - buffers across chunks
  - supports CRLF (`\r\n`)
  - supports multiline `data:` (joined with `\n` per SSE spec)
  - uses strict SSE framing (does not flush a partial final block at EOF)
- Migrated:
  - `openAIAdapter()` (Chat Completions SSE)
  - `openAIResponsesAdapter()` (Responses API SSE)
  to consume the shared decoder instead of doing ad-hoc line splitting.
- Added focused tests:
  - unit tests for the SSE decoder
  - regression tests proving OpenAI SSE events can be split across chunks without being dropped

## Why it matters
This hardens streaming behavior under normal network conditions and reduces duplicated parsing logic across adapters—making streaming output less flaky and easier to maintain.

## Scope note
`agUIAdapter()` is intentionally unchanged in this PR (it already has robust SSE framing); this change focuses on bringing the OpenAI SSE adapters up to the same standard.

## How to test
```bash
corepack pnpm --filter @openuidev/react-headless test

### DIFF
--- a/packages/react-headless/src/stream/sse/decodeSSE.ts
+++ b/packages/react-headless/src/stream/sse/decodeSSE.ts
@@ -1,54 +1,64 @@
+import { createParser, type ParsedEvent } from "eventsource-parser";
+
 export type SSEEvent = {
   data: string;
 };
 
-const normalizeEols = (chunk: string) => chunk.replace(/\r\n/g, "\n");
-
-const stripOptionalLeadingSpace = (value: string) => (value.startsWith(" ") ? value.slice(1) : value);
-
-const parseSSEBlockData = (block: string): string => {
-  const dataLines: string[] = [];
-
-  for (const rawLine of block.split("\n")) {
-    if (!rawLine) continue;
-    if (rawLine.startsWith(":")) continue; // comment / keepalive
-
-    if (rawLine.startsWith("data:")) {
-      dataLines.push(stripOptionalLeadingSpace(rawLine.slice(5)));
-    }
-  }
-
-  return dataLines.join("\n");
-};
-
 /**
- * Minimal SSE decoder:
+ * SSE decoder using eventsource-parser:
  * - buffers across chunks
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by abhithesys — COMMENTED** [review]
Changes look good to me. Can you replace the decodeSSE implementation with [eventsource-parser](https://www.npmjs.com/package/eventsource-parser). It's battle tested and we don't need to maintain the decoding code. 

Thanks for simplifying this @boffin-dmytro.

File: packages/react-headless/src/stream/sse/decodeSSE.ts [hunk 1]
```diff
@@ -0,0 +1,64 @@
+import { createParser, type ParsedEvent } from "eventsource-parser";
```
abhithesys, line 1: this file is too complicated. can you try doing something like this:

```
import { EventSourceParserStream } from "eventsource-parser/stream";

export type SSEEvent = {
  data: string;
};
t async function* decodeSSE(response: Response): AsyncIterable<SSEEvent> {
  const body = response.body;
  if (!body) throw new Error("No response body");

  const eventStream = body
    .pipeThrough(new TextDecoderStream())
    .pipeThrough(new EventSourceParserStream());

  for await (const event of eventStream) {
    yield { data: event.data };
  }
}
```

PR #350, opened 2026-03-17
feat(react-lang): expand parser validation with rule-tagged errors

### LINKED ISSUE #349: Improve Parser Validation
**Describe the bug**

`createParser` and `createStreamingParser` do not report errors for invalid OpenUI Lang output. The parser silences errored output with no signal, making it hard to detect hallucinated model responses or incomplete streams.

Three specific cases:

- Unknown component names: If the model emits a component name not in the library schema, the component is rendered with its arguments preserved under _args, and no error is reported.
- If the model passes more arguments than a component's schema defines, the extra args are silently dropped. No error is reported.
- If the model references an identifier that is never assigned, the reference is dropped as null and tracked in `meta.unresolved`, but never promoted to meta.validationErrors. After the stream completes, there is no error.

**Expected behavior**

Each case should produce a structured entry in meta.validationErrors with enough context to identify the problem. Errors should not affect rendering (the parser should stay permissive), but they should be observable.

For unresolved references specifically: mid-stream forward refs are expected and should not error. Only after the stream is declared complete should still-unresolved refs become errors.

**Screenshots**

Image

row1, row2, row3 are all undefined in the response, and it is also passing more args than present in the signature. This is while using Gemini 2.5 Flash with a modified version of the `openuidev/cli` generated system prompt. This is leading to blank visual responses despite the streaming having been finished. Definitely requires better prompt engineering by the consumer, but would be helpful for downstream library to validate and report.

**Additional context**

Happy to share the link to the private repo for additional context

AbhinRustagi:
## Summary

Closes #349

- Adds three new validation checks to `ParseResult.meta.errors`: `unknown-component`, `excess-args`, and the existing `missing-required`/`null-required` rules
- Replaces `ValidationError`/`ValidationRule` with a discriminated `OpenUIError` type using `type` + `code` fields — extensible for future `type: "parser"` diagnostics
- Renames `meta.validationErrors` → `meta.errors`
- Removes `StreamParser.finalize()` — unresolved references stay in `meta.unresolved` for consumers to check after streaming ends
- Exports new `OpenUIError` and `ValidationErrorCode` types from the package index

Rendering behavior is unchanged — the parser stays permissive and renders what it can.

## Test plan

- [x] `pnpm test` in `packages/react-lang` — 16 tests, all passing
- [x] `unknown-component`: parse `root = DataTable("col")` with `DataTable` absent from schema → `errors` contains `{ type: "validation", code: "unknown-component" }`
- [x] `excess-args`: parse `root = Title("a", "b")` where `Title` takes 1 arg → `errors` contains `{ type: "validation", code: "excess-args" }`
- [x] Unresolved refs tracked in `meta.unresolved` without producing errors (both one-shot and streaming)
- [x] Streaming: unresolved refs resolve automatically when defined in later chunks
- [x] Existing `missing-required` and `null-required` errors carry new `type`/`code` shape
- [x] TypeScript compiles cleanly (`tsc --noEmit`)


AbhinRustagi:
@Aditya-thesys Okay, I will update the PR to remove `.finalize()`, keep `.meta.unresolved` as is, and update the error signature to a generalized `OpenUIError`. Do you reckon I should update the docs for now to address the current error monitoring methodologies, unless OpenUI Lang 1.5 is close to completion?

AbhinRustagi:
I was thinking of adding a Guides section to [OpenUI page](https://www.openui.com/docs/openui-lang) because the error handling is supposed to be suggestive and a how-to of how the library offers visibility on errors, and not a strict mandate.

AbhinRustagi:
@Aditya-thesys Updated the PR to the latest discussion

PR #403, opened 2026-03-26
feat(cli): prompt for OpenAI API key in interactive mode

### LINKED ISSUE #399: Ask for OpenAI key in interactive mode
Right now at the end of the installation step, the cli prints this:
```bash
Done!
Get started:

cd foo

touch .env

Add your API key to .env:
OPENAI_API_KEY=sk-your-key-here

pnpm run dev
```
It would be better to ask for the key in interactive mode, create the .env file ourself and start the server directly

In interactive mode it would be better to get the server up and running and give users an url to start working with

**Additional context**
- In non interactive mode skip this step entirely. 
- No need for --openai-key=<> flag for now

AliaksandrNazaruk:
## Summary

When running `create-chat-app` interactively, prompt the user for their OpenAI API key instead of printing manual instructions. If a key is provided, the CLI creates the `.env` file automatically and simplifies the getting-started output. In non-interactive mode, behavior is unchanged.

## Changes

- **`packages/openui-cli/src/commands/create-chat-app.ts`**: After installing dependencies, prompt for the OpenAI API key using the existing `resolveArgs` helper. If provided, write `.env` and show a streamlined message. If skipped (blank input) or non-interactive, show the original instructions.

Closes #399

### DIFF
--- a/packages/openui-cli/src/commands/create-chat-app.ts
+++ b/packages/openui-cli/src/commands/create-chat-app.ts
@@ -117,9 +117,7 @@ export async function runCreateChatApp(options: CreateChatAppOptions): Promise<v
     ).trim();
 
     if (apiKey) {
-      const envPath = path.join(targetDir, ".env");
-      fs.writeFileSync(envPath, `OPENAI_API_KEY=${apiKey}\n`);
-      console.info("\n✅ .env file created with your API key.\n");
+      fs.writeFileSync(path.join(targetDir, ".env"), `OPENAI_API_KEY=${apiKey}\n`);
       apiKeyWritten = true;
     }
   }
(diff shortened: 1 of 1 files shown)


packages/openui-cli/src/commands/create-chat-app.ts
```diff
@@ -102,7 +102,33 @@ export async function runCreateChatApp(options: CreateChatAppOptions): Promise<v
           ? "bun"
           : "npm";
 
-  console.info(getStartedMessage(name, devCmd));
+  if (!options.noInteractive) {
+    const apiKeyArgs = await resolveArgs(
```
rabisg: resolveOne was meant as a utility for when the value can be passed from both the cli as flags and in interactive mode. Since we don't want this in non-interactive mode or a flag for this, you can directly use inquirer here

AbhinRustagi: +1

Also, this can be reorganized into a fn in the `lib/` folder to clean up the command and keep it as a high-level orchestrator.

AbhinRustagi: There's also a merge conflict to be resolved with the skill option. 

I think it can be extended to a 3-step scenario

- Ask for OpenAI SDK BaseUrl (read a comment/issue about someone using OpenRouter, I personally used Gemini) (default value can be something like "Use OpenAI")
- (If they add a custom url) Ask for model name
- Ask for Api Key

Wdyt?

Drawbacks: prolongs the setup process

rabisg: We can keep this PR simple and just take the API key
We can discuss the long term roadmap for OpenUI cli setup on the issue

PR #426, opened 2026-04-06
Add a flag to bootstrap a new project from an example instead of blank template

rabisg:
## What

Adds an --example flag to the openui create command so users can scaffold directly from one of the repo's example apps instead of only the built-in default template. In interactive mode, users are prompted to choose between the clean template and an example before being shown a searchable list of options.

## Changes
- Added --example <name> and --list-examples flags to the create command
- Added interactive "Start from: Clean template / Example from the repo" prompt when neither flag is passed
- Added src/lib/fetch-example.ts — downloads the example from the GitHub tarball at scaffold time (no git required on the user's machine)
- Added scripts/gen-examples-list.mjs - reads the examples/ directory at build time and generates src/generated/known-examples.ts so the example list is always in sync without manual maintenance
- Wired build:examples-list as the first step in the build script so it runs automatically
- Added packages/openui-cli/src/generated/ to .gitignore

## Test Plan

- [x] Verified locally

## Checklist


packages/openui-cli/src/commands/create-chat-app.ts
```diff
@@ -43,39 +96,45 @@ export async function runCreateChatApp(options: CreateChatAppOptions): Promise<v
...
+    const pkg = JSON.parse(fs.readFileSync(pkgPath, "utf8")) as {
+      name: string;
+      dependencies?: Record<string, string>;
+      devDependencies?: Record<string, string>;
+    };
+    pkg.name = name;
+    for (const section of ["dependencies", "devDependencies"] as const) {
+      for (const key of Object.keys(pkg[section] ?? {})) {
+        if (pkg[section]![key] === "workspace:*") {
+          pkg[section]![key] = "latest";
```
abhithesys, line 132: can we update this to use latest version number fetched from npm?

ChinmayMhatre:
In the newer changes, There is a non nullable template flag for openui cloud and self hosted.
Would it make sense to add these in the `-t` flag?
create-chat-app has also been reworked to create-app

PR #440, opened 2026-04-08
shell-updates

pr3khar:
Shell parity and layout (Shell, Bottom tray, Copilot Shell)
Aligned behavior and visuals across standalone Shell, BottomTray, and CopilotShell: containers, headers, threads, thread lists, welcome screens, conversation starters, and composers.

CopilotShell: added a ThreadListContainer plus threadList.scss so its thread list matches the other layouts more closely.

API and state:
SharedChatUIProps / OpenUIChat: new optional showAssistantLogo; composed layouts pass it through.

useScrollToBottom: respects scroll-padding-top when scrolling to the last user message, and clamps scroll position with Math.max(..., 0) to avoid odd negative scrolls.


ankit-thesys (approved):
Looks good

PR #476, opened 2026-04-24
fix(a11y): add aria-labels to chart scroll buttons

akinshaywai:
## Summary

The chart scroll buttons (horizontal left/right and vertical up/down) used icon-only `IconButton` components with no accessible name, failing [WCAG 2.1 SC 4.1.2 (Name, Role, Value)](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html).

Screen readers would announce these as unlabelled buttons, giving users no indication of what they do.

**Files changed:**
- `Charts/shared/ScrollButtonsHorizontal/ScrollButtonsHorizontal.tsx` — added `aria-label="Scroll left"` and `aria-label="Scroll right"`
- `Charts/shared/ScrollButtonsVertical/ScrollButtonsVertical.tsx` — added `aria-label="Scroll up"` and `aria-label="Scroll down"`

## Test plan

- [ ] Open a chart with overflow content using a screen reader (VoiceOver / NVDA)
- [ ] Verify the scroll buttons announce "Scroll left", "Scroll right", "Scroll up", "Scroll down" respectively


ankit-thesys (approved):
Looks good

PR #482, opened 2026-04-28
typography updates and semicircle fix

pr3khar:
added number-heading variables and semicircle fixes for pie/donut


ankit-thesys (approved):
LG

PR #485, opened 2026-05-02
refactor(react-ui): simplify Accordion auto-open logic

i-subham23:
Updated the Accordion component to streamline the auto-open functionality. Removed the previous content size tracking and replaced it with a simpler check that opens the newest item when a new item arrives during streaming. This enhances performance and reduces complexity in state management.

https://github.com/user-attachments/assets/f985eb30-c87f-4452-8038-2fb023097447


ankit-thesys (approved):
Looks good

PR #487, opened 2026-05-04
Guard chart components against null streaming data

### LINKED ISSUE #355: Charts component crashes with TypeError during streaming: null data passed to Recharts
## Bug Description

The built-in `Charts` component in `@openuidev/react-ui` throws a `TypeError: Cannot read properties of null (reading 'map')` during streaming rendering. The error originates from Recharts' `CategoricalChartWrapper.render()` when the OpenUI `Renderer` passes partial/incomplete data to the chart component before the full dataset has streamed in.

## Steps to Reproduce

1. Scaffold a new project: `npx @openuidev/cli@latest create --name test-app`
2. Configure an LLM provider (OpenRouter with Claude Sonnet 4 in my case)
3. Send a prompt that generates a chart, e.g.: *"Show me a CEO dashboard with quarterly revenue data for 4 regions"*
4. Observe the console error during streaming

## Error Stack Trace

```
TypeError: Cannot read properties of null (reading 'map')
    at CategoricalChartWrapper.render (recharts/es6/chart/...)
    at renderByOrder (recharts/es6/util/...)
    at RenderNode (@openuidev/...)
    at renderDeep (@openuidev/...)
    at Charts component (@openuidev/react-ui/dist/genui-lib/...)
    at GenUIAssistantMessage (...)
```

## Expected Behavior

The Charts component should guard against null/undefined data during progressive streaming. A simple `data ?? []` fallback before passing to Recharts would prevent this.

## Actual Behavior

The chart eventually renders correctly after streaming completes, but the console error fires during the streaming phase when the chart structure arrives before its data array.

## Environment

- `@openuidev/react-ui`: latest (installed 2026-03-18)
- `@openuidev/react-lang`: latest
- `@openuidev/react-headless`: latest
- Next.js 16.1.6 (Turbopack)
- React 19.2.3
- LLM: Claude Sonnet 4 via OpenRouter
- OS: Windows 11

## Suggested Fix

In the Charts genui-lib component, add a null guard before passing data to Recharts:

```typescript
// Before passing to <BarChart data={data}>
const safeData = data ?? [];
```

## Additional Note

The `FullScreen` component also defaults `logoUrl` to `https://crayonai.org/img/logo.png` which appears to be a dead URL from the Crayon AI rebrand, causing broken image icons in the header and assistant avatar. Consider updating the default or removing it.

CodexNexor:
## Summary
- add a shared `ensureChartData()` helper that normalizes `null`/`undefined` chart data to `[]`
- apply the runtime guard across the GenUI-facing chart components that can receive partial streaming props before the full payload arrives
- keep export/legend/domain calculations aligned with the guarded data so Recharts never receives a null dataset during streaming

## Why this fix
The chart schemas in current `main` already require the expected fields, but that still doesn't fully protect the runtime streaming path when partial data reaches the components before the payload is complete. This PR adds the missing runtime safety layer without changing the public API.

## Validation
- ran `git diff --check`
- manually verified the guarded components now derive their Recharts `data` prop from `ensureChartData(...)`

Closes #355.

### DIFF
--- a/packages/react-ui/src/components/Charts/AreaChartCondensed/AreaChartCondensed.tsx
+++ b/packages/react-ui/src/components/Charts/AreaChartCondensed/AreaChartCondensed.tsx
@@ -27,6 +27,7 @@ import { LabelTooltipProvider } from "../shared/LabelTooltip/LabelTooltip";
 import { LegendItem } from "../types";
 import { getLineType } from "../utils/AreaAndLine/common";
 import {
+  ensureChartData,
   get2dChartConfig,
   getColorForDataKey,
   getDataKeys,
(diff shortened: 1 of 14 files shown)


ankit-thesys (commented):
Changes look good. Can you fix the build issue? You are good to go with this PR.

Sh-dheeraj:
Fixed the build issue and pushed an update to this PR.

Root cause was a Prettier failure in two chart files touched by this change.

Latest commit: `67077e3` (`Fix formatting in chart null guard changes`)

Please recheck when you get a chance.

ankit-thesys:
Just read the issue. Have you tried the solution suggested there?
I think this is not required in the component library; instead, it should be handled by the renderer or consumer.

Sh-dheeraj:
Updated this PR to follow your suggestion.

I moved the runtime guard out of the chart component library and into `packages/react-ui/src/genui-lib`, so incomplete streamed props now short-circuit in the renderer layer before chart data reaches the components.

What changed in this update:
- reverted the `components/Charts/*` null-guard changes from this PR
- added stricter streamed-prop normalization in `genui-lib/helpers.ts`
- reused those helpers in the chart wrappers for array charts, categorical charts, and scatter datasets
- kept legacy slice fallback behavior intact

Latest commit: `556a592` (`Move chart null guards into genui renderer`)

Validation on my side:
- `git diff --check`
- targeted Prettier check on the touched files using the repo's Prettier settings/plugin

I couldn't run the full workspace typecheck/build here because this environment doesn't have the repo toolchain installed, but the PR is now aligned with the renderer/consumer approach you mentioned. Please recheck when you get a chance.

ankit-thesys:
Hey! First of all, thank you for taking the time to put this PR together — we genuinely appreciate the contribution and the effort that went into it.

After carefully reviewing the PR and discussing it with the React Renderer maintainers, we’ve decided to address this by adding guards for nullish data on the renderer side in the react-lang package. Since that approach resolves the issue at the source, we won’t be moving forward with this PR.

That said, your contribution was genuinely helpful — it surfaced the issue clearly and gave us the push to investigate and fix it in what we believe is the right place. We really appreciate you taking the initiative here, and hope you’ll continue contributing in the future.

PR #538, opened 2026-05-21
fix: add local SVG icons for xAI and DeepSeek providers (Closes #537)

### LINKED ISSUE #537: Fix missing xAI and DeepSeek provider icons on homepage
On the OpenUI homepage, two LLM provider icons in the compatibility section are not rendering correctly.

In the “Works effortlessly with your favorite stack” section, the xAI and DeepSeek chips show broken/missing image icons instead of the expected provider logos.

**Affected area:**

`docs/app/(home)/sections/CompatibilitySection/CompatibilitySection.tsx`

**Observed behavior:**

- The xAI chip renders with a missing image icon.
- The DeepSeek chip renders with a missing image icon.
- Other provider icons like OpenAI, Anthropic, Gemini, and Mistral appear correctly.

**Expected behavior:**

xAI and DeepSeek should display their provider logos consistently with the other LLM provider chips.

**Possible cause:**

The `slug` values or image source resolution for these two providers may not map to valid assets in the icon provider being used.

**Suggested fix:**

- Verify whether `xai` and `deepseek` resolve to valid provider icon assets.
- If not, add local SVG assets for these providers or update the slug values to the correct supported identifiers.
- Confirm the homepage compatibility section renders all LLM provider icons correctly after the fix.

Image

magnet78:
## Issue
Closes #537

## Summary
The xAI and DeepSeek provider icons in the homepage compatibility section were broken because their `slug` values (`xai` and `deepseek`) don't exist on the simple-icons CDN. This PR fixes the issue by adding local SVG icon assets.

## Changes
- Added `docs/public/brand-icons/xai.svg` — xAI brand icon
- Added `docs/public/brand-icons/deepseek.svg` — DeepSeek brand icon
- Updated `CompatibilitySection.tsx` to use `localSrc` instead of `slug` for both providers

## Why this approach
Other providers (Anthropic, Mistral, Gemini) use `slug` because they exist on simple-icons.org CDN. OpenAI already uses `localSrc` with a local SVG. This fix follows the same pattern as OpenAI for the two providers not available on simple-icons.


vishxrad (commented):
I do not think this is the right fix as-is.

Two concerns:

1. `deepseek` does exist as a Simple Icons slug, so we should not replace it with a local asset unless there is another reason. The newly added `deepseek.svg` also does not look like the DeepSeek brand icon.

2. `xai` does not appear to exist as a Simple Icons slug, but the newly added `xai.svg` looks like a generic Vercel-style triangle/mark, not the xAI brand icon.

So I think we should either:

- keep DeepSeek on the Simple Icons CDN with `slug: "deepseek"`, and
- source an accurate xAI SVG from an official/approved source, or use a text fallback for xAI until we have the right asset.

The local asset pattern is fine in principle, since OpenAI already uses `localSrc`, but the icons added here do not look accurate.

PR #572, opened 2026-05-25
feat(react-lang): support custom host wrappers for React Native

### LINKED ISSUE #523: React Native support for react-lang Renderer needs native host wrappers
## Summary

The react-lang Renderer is difficult to use directly inside React Native and Expo because the default implementation assumes web host elements.

## What happened

While migrating the React Native example to OpenUI Lang v0.5 and @openuidev/react-lang 0.2.5, Android Expo Go crashed with:

`View config getter callback for component div must be a function received undefined`

The crash came from Renderer using DOM host wrappers by default. React Native treats lowercase host names such as div as native component names, so the renderer cannot be dropped into a native app without a way to replace those wrappers with View.

A second integration issue showed up when linking the workspace package into Expo. Metro could resolve react-lang from the workspace, but react-lang could then resolve its own dev React instead of the app peer React. That produced invalid hook call errors until the Expo Metro config forced React imports back to the app copy.

## Expected behavior

React Native hosts should be able to use the official react-lang Renderer without maintaining a separate renderer implementation.

## Proposed fixes

- Provide native safe host wrapper configuration on Renderer, for example containerComponent and contentComponent.
- Provide a native compatible query loader path, or let hosts fully replace it.
- Document Expo and pnpm workspace peer React resolution requirements.
- Consider a react-native entrypoint that defaults wrappers to View instead of div.

## Minimal usage goal

```tsx
<Renderer
  response={openui}
  library={library}
  isStreaming={isStreaming}
  toolProvider={toolProvider}
  containerComponent={View}
  contentComponent={View}
/>
```

## Impact

Without this, React Native examples either crash on native devices or need an example-local renderer, which makes actions, Query refresh, streaming behavior, and future react-lang runtime fixes easier to drift from the official renderer.

greatjourney589:
## What
Closes #523 
Add support for custom host wrappers in `@openui/react-lang`'s `Renderer` so it can render on non-web hosts like React Native / Expo.

Today the `Renderer` hardcodes `<div>` for its outer and inner wrappers and always falls back to a DOM spinner. On React Native, lowercase host tags like `"div"` are not valid components and crash with *"View config getter callback for component `div` ..."*. This change lets callers supply native host components (e.g. `View` from `react-native`) instead.

## Changes

- Add `containerComponent` and `contentComponent` props to `RendererProps` (default to `"div"`), so the outer/inner wrappers can be swapped for native hosts.
- Apply the web-only CSS `opacity` fade transition only when the content wrapper is a web host (string tag); skip it for native components to avoid warnings/crashes.
- Fall back to the default DOM spinner only on web hosts; on native, render nothing unless the caller passes their own `queryLoader`.
- Add `react-dom` + `@types/react-dom` as devDependencies and a `vitest.config.ts` to support component tests.
- Add `Renderer` tests covering default web `div` wrappers, custom native wrappers, transition omission on native, and null-response rendering.
- Update `packages/react-lang/README.md` with React Native usage docs.

## Test Plan

- [ ] Not applicable (explain why)
- [x] Verified locally — `vitest run` in `packages/react-lang` passes (4/4 tests), covering web defaults, native wrappers, web-only transition handling, and null responses.

## Checklist

- [x] I considered backwards compatibility — new props are optional and default to `"div"`, so existing web usage is unchanged.

### DIFF
--- a/packages/react-lang/README.md
+++ b/packages/react-lang/README.md
@@ -42,9 +42,7 @@ const Greeting = defineComponent({
     mood: z.enum(["happy", "excited"]).optional().describe("Tone of the greeting"),
   }),
   component: ({ name, mood }) => (
-    <div className={mood === "excited" ? "text-xl font-bold" : ""}>
-      Hello, {name}!
-    </div>
+    <div className={mood === "excited" ? "text-xl font-bold" : ""}>Hello, {name}!</div>
   ),
 });
 ```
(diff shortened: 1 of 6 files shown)


greatjourney589:
Hi, @vishxrad 
I am very interested in your repository.This is my first PR.
When you have a moment, could you please review this PR?

PR #580, opened 2026-05-26
fix(chat): replace unsafe new Function() with math expression parser

sebastionoss:
## What

Fix a code injection vulnerability (CWE-94) in the `calculate()` tool function in `docs/app/api/chat/route.ts`. The existing implementation uses `new Function()` to evaluate math expressions with a regex-based sanitizer that is trivially bypassable, allowing arbitrary JavaScript execution on the server.

## Changes

- Replaced the `new Function(`return (${sanitized})`)()` call with a safe recursive-descent math expression parser
- The new parser only supports numeric literals, arithmetic operators (`+`, `-`, `*`, `/`, `%`), parentheses, and a whitelist of `Math` functions (`sqrt`, `pow`, `abs`, `ceil`, `floor`, `round`)
- Removed the regex-based sanitizer entirely since the parser rejects anything that isn't valid math

## Vulnerability Details

The original code sanitizes user-supplied math expressions with this regex:

```ts
const sanitized = expression.replace(
  /[^0-9+\-*/().%\s,Math.sqrtpowabsceilfloorround]/g,
  "",
);
const result = new Function(`return (${sanitized})`)();
```

The regex uses a **character class** — `[^...Math.sqrtpowabsceilfloorround]` — which permits individual characters `M`, `a`, `t`, `h`, `s`, `q`, `r`, `p`, `o`, `w`, `b`, `c`, `e`, `i`, `l`, `f`, `d`, `u`, `n`, not the literal strings `Math.sqrt` etc. This means the characters needed to spell `constructor`, `this`, `process`, `return`, and other JavaScript identifiers all pass through the filter.

An attacker who can influence the `expression` parameter passed to the `calculate` tool can execute arbitrary code on the server. The `calculate` function is registered as a tool for the LLM chat endpoint (`POST /api/chat`), and the LLM can be prompted to invoke it with a crafted expression.

### Proof of Concept

The following expression passes the regex sanitizer completely unchanged (every character is in the allowed set):

```
constructor.constructor("return process.cwd()")()
```

Breaking down why this passes:
- `c`, `o`, `n`, `s`, `t`, `r`, `u` are all individually in the character class (from `constructor` letters overlapping with `sqrtpowabsceilfloorround`)
- `.` is explicitly allowed
- `(`, `)` are allowed
- `"` would be stripped, but the attack can use backticks or other JS syntax tricks

A more direct payload:

```
(()=>{return this.constructor.constructor("return process")()})()
```

While some characters get stripped, the core identifiers `constructor`, `process`, `return`, `this` survive because their constituent characters are all in the allowed set. To reproduce, paste into Node.js:

```js
const expression = 'constructor.constructor("return process.cwd()")()';
const sanitized = expression.replace(
  /[^0-9+\-*/().%\s,Math.sqrtpowabsceilfloorround]/g,
  "",
);
console.log("Sanitized:", sanitized);
// Many characters survive, enabling code construction
const result = new Function(`return (${sanitized})`)();
console.log("Result:", result);
```

### Why existing mitigations don't prevent this

Before submitting, we verified that the vulnerability is not mitigated by other controls. The `/api/chat` endpoint is a Next.js API route with no middleware-level authentication — there is no `middleware.ts` at the docs app root, and no auth checks in the route handler itself. The regex sanitizer is the only defense, and as shown above, it does not block code injection payloads because it operates on individual characters rather than keywords. The `setTimeout` wrapper and `try/catch` do not prevent execution — they simply defer it.

## Fix Rationale

Rather than attempting to improve the regex (which is fundamentally the wrong approach for code injection prevention), this fix replaces the entire evaluation mechanism with a recursive-descent parser that:

1. **Tokenizes** the input into numbers, operators, function names, parentheses, and commas
2. **Validates** function names against a strict allowlist before accepting them as tokens
3. **Evaluates** using standard operator precedence (unary → multiplication/division → addition/subtraction)
4. **Rejects** any character or identifier not in the grammar with a clear error

This approach cannot execute arbitrary code because it never calls `eval()`, `new Function()`, or any other code execution primitive.

## Test Plan

- [x] Verified locally

Tested the parser with:
- Basic arithmetic: `2 + 3`, `10 * 5 / 2`, `(3 + 4) * 2`
- Math functions: `sqrt(16)`, `Math.pow(2, 8)`, `abs(-5)`, `ceil(3.2)`
- Unary operators: `-5 + 3`, `+10`
- Nested expressions: `sqrt(pow(3, 2) + pow(4, 2))`
- Malicious inputs: `constructor`, `process`, `require("child_process")` — all correctly rejected with "Unknown identifier" or "Invalid character" errors

## Checklist

The fix is backwards-compatible — all valid math expressions that worked before continue to work. Only malicious/non-math inputs are now rejected instead of executed.

### DIFF
--- a/docs/app/api/chat/route.ts
+++ b/docs/app/api/chat/route.ts
@@ -105,15 +105,192 @@ function getStockPrice({ symbol }: { symbol: string }): Promise<string> {
   });
 }
 
+// ── Safe math expression parser (replaces new Function) ──
+
+type MathToken =
+  | { type: "number"; value: number }
+  | { type: "op"; value: string }
+  | { type: "func"; value: string }
+  | { type: "paren"; value: string }
+  | { type: "comma" };
+
+const ALLOWED_FUNCTIONS = new Set([
+  "Math.sqrt",
+  "Math.pow",
+  "Math.abs",
+  "Math.ceil",
+  "Math.floor",
+  "Math.round",
+  "sqrt",
+  "pow",
+  "abs",
+  "ceil",
+  "floor",
+  "round",
+]);
+
+const MATH_FNS: Record<string, (...args: number[]) => number> = {
+  sqrt: Math.sqrt,
+  pow: Math.pow,
+  abs: Math.abs,
+  ceil: Math.ceil,
+  floor: Math.floor,
+  round: Math.round,
+};
+
+function tokenizeMath(expr: string): MathToken[] {
+  const tokens: MathToken[] = [];
+  let i = 0;
+  while (i < expr.length) {
+    const ch = expr[i];
+    if (/\s/.test(ch)) {
+      i++;
+      continue;
+    }
+    if (/[0-9]/.test(ch) || (ch === "." && i + 1 < expr.length && /[0-9]/.test(expr[i + 1]))) {
+      let num = "";
+      while (i < expr.length && (/[0-9]/.test(expr[i]) || expr[i] === ".")) num += expr[i++];
+      tokens.push({ type: "number", value: parseFloat(num) });
+      continue;
+    }
+    if ("+-*/%".includes(ch)) {
+      tokens.push({ type: "op", value: ch });
+      i++;
+      continue;
+    }
+    if (ch === "(" || ch === ")") {
+      tokens.push({ type: "paren", value: ch });
+      i++;
+      continue;
+    }
+    if (ch === ",") {
+      tokens.push({ type: "comma" });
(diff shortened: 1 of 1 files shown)


vishxrad:
Thanks a lot @sebastiondev for pointing this out! We will review it and get back to you ASAP.

vishxrad:
Thanks a lot for raising this issue, I will close this pr now since #600 closes it.

PR #586, opened 2026-05-30
fix(react-headless): fall back when crypto.randomUUID is unavailable

Shinyaigeek:
## Problem

Sending a message throws a runtime `TypeError: crypto.randomUUID is not a function` at `packages/react-headless/src/store/createChatStore.ts` (optimistic message id generation).

`crypto.randomUUID()` is only available in a **secure context** — HTTPS, or `http://localhost` / `127.0.0.1` / `[::1]`. When an app built on `@openuidev/react-headless` is opened from a non-secure origin (a LAN IP over plain HTTP, `0.0.0.0`, an internal hostname, etc.), `crypto.randomUUID` is `undefined` and the call throws.

## Fix

Add a small `safeRandomUUID()` helper that uses `crypto.randomUUID()` when available and falls back to a sufficiently-unique id otherwise (adequate for client-side message keys), and use it wherever the store generated ids.

## Notes

- No behavior change in secure contexts.
- `typecheck` + `build` pass for `@openuidev/react-headless`.


Shinyaigeek:
Sorry this is ai slop 🙏

PR #599, opened 2026-06-04
feat: Add native Swift integration for OpenUI (#393)

### LINKED ISSUE #393: Native Swift Integration
OpenUI currently focuses on JavaScript/TypeScript runtimes and examples. Teams building native iOS, iPadOS, or macOS apps with Swift do not have a native way to define OpenUI component libraries, generate OpenUI Lang prompts, parse streamed OpenUI Lang output, or render generated UI directly in SwiftUI.

This means Apple-platform integrations would likely need to rely on a WebView, a bridge to the web runtime, or a custom app-specific parser and renderer. That adds friction for Swift-first teams and makes OpenUI harder to adopt in native Apple apps.

Add native Swift support for OpenUI, ideally targeting SwiftUI first.

The integration should make it possible to:

- Define OpenUI component libraries in Swift with typed props.
- Generate system prompts from those component definitions.
- Parse OpenUI Lang output, including streamed/incremental output.
- Render parsed OpenUI Lang into SwiftUI views.
- Support common OpenUI primitives such as layout, text, forms, buttons, lists, tables, and charts where feasible.
- Expose action/event callbacks so generated UI can trigger app behavior.
- Provide a minimal SwiftUI sample app showing an end-to-end streaming chat flow.

A good first version could focus on the core Swift runtime and SwiftUI renderer, with parity for the most common OpenUI Lang features before expanding to the full component surface.

**Describe alternatives you've considered**
- Use a WebView and render the existing React runtime inside a native Apple app.
- Keep OpenUI rendering on the server and send pre-rendered UI to the app.
- Build a custom app-specific Swift parser for a limited subset of OpenUI Lang.
- Use React Native instead of SwiftUI for mobile OpenUI experiences.

These options work around the gap, but they either reduce native integration quality or create duplicated protocol/rendering work.

**Additional context**
This would make OpenUI more useful for teams building native Apple-platform apps with Swift and SwiftUI. It would also help position OpenUI Lang as a portable UI generation protocol that can target native mobile and desktop frameworks, not only web runtimes.

Potential package/module shape:

- `OpenUILang` for parsing, validation, prompt generation, and streaming state.
- `OpenUISwiftUI` for SwiftUI rendering.
- `Examples/SwiftUIChat` for an end-to-end demo.

amankoli09:
Closes #393 

+What this PR does / why we need it:
This PR introduces native Swift integration for OpenUI, bringing the power of the openui-lang protocol to Apple-platform native apps without requiring a WebView.

This provides Swift-first teams the tools to define components, generate prompts, parse streamed outputs, and render them natively using SwiftUI.

**Changes included in this PR:**
1. **`OpenUILang` (Core Swift Package)**
   - `AST.swift` & `Types.swift`: Native data models mapping to the OpenUI specification (`ASTNode`, `ElementNode`, `ParseResult`).
   - `Library.swift`: A type-safe registry (`ComponentLibrary` and `ComponentDef`) to define OpenUI component libraries natively in Swift.
   - `PromptGenerator.swift`: Logic to compile the `ComponentLibrary` into a system prompt.
   - `Parser.swift`: A native Swift parser designed to handle OpenUI Lang streaming text formats and compile them into a reactive AST.

2. **`OpenUISwiftUI` (Renderer Package)**
   - `OpenUIRenderer.swift`: A dynamic, recursive SwiftUI `View` that traverses the `ElementNode` tree and maps it to native UI elements (currently supports `VStack`, `HStack`, `Text`, and `Button`). It uses `AnyView` to handle recursive structural cycles in SwiftUI.

3. **`SwiftUIChat` (Example Demo App)**
   - A standalone macOS/iOS SwiftUI app showcasing an end-to-end streaming chat integration.

### DIFF
--- a/packages/swift/.gitignore
+++ b/packages/swift/.gitignore
@@ -1,0 +1,8 @@
+.DS_Store
+/.build
+/Packages
+xcuserdata/
+DerivedData/
+.swiftpm/configuration/registries.json
+.swiftpm/xcode/package.xcworkspace/contents.xcworkspacedata
+.netrc
--- a/packages/swift/Package.swift
+++ b/packages/swift/Package.swift
@@ -31,6 +31,9 @@ let package = Package(
             dependencies: ["OpenUILang"]),
         .executableTarget(
             name: "SwiftUIChat",
-            dependencies: ["OpenUILang", "OpenUISwiftUI"])
+            dependencies: ["OpenUILang", "OpenUISwiftUI"]),
+        .testTarget(
+            name: "OpenUILangTests",
+            dependencies: ["OpenUILang"])
     ]
 )
--- a/packages/swift/Package.swift
+++ b/packages/swift/Package.swift
@@ -1,0 +1,36 @@
+// swift-tools-version: 5.9
+// The swift-tools-version declares the minimum version of Swift required to build this package.
+
+import PackageDescription
+
+let package = Package(
+    name: "OpenUI",
+    platforms: [
+        .macOS(.v13),
+        .iOS(.v16)
+    ],
+    products: [
+        // Products define the executables and libraries a package produces, making them visible to other packages.
+        .library(
+            name: "OpenUILang",
+            targets: ["OpenUILang"]),
+        .library(
+            name: "OpenUISwiftUI",
+            targets: ["OpenUISwiftUI"]),
+        .executable(
+            name: "SwiftUIChat",
+            targets: ["SwiftUIChat"])
+    ],
+    targets: [
+        // Targets are the basic building blocks of a package, defining a module or a test suite.
+        // Targets can depend on other targets in this package and products from dependencies.
+        .target(
+            name: "OpenUILang"),
+        .target(
+            name: "OpenUISwiftUI",
+            dependencies: ["OpenUILang"]),
+        .executableTarget(
+            name: "SwiftUIChat",
+            dependencies: ["OpenUILang", "OpenUISwiftUI"])
+    ]
+)
--- a/packages/swift/Sources/OpenUILang/AST.swift
+++ b/packages/swift/Sources/OpenUILang/AST.swift
@@ -1,0 +1,40 @@
+public indirect enum ASTNode {
+    case comp(name: String, args: [ASTNode], mappedProps: [String: ASTNode]?)
+    case str(String)
+    case num(Double)
+    case bool(Bool)
+    case null
+    case arr([ASTNode])
+    case obj([(String, ASTNode)])
+    case ref(String)
+    case ph(String)
+    case stateRef(String)
+    case runtimeRef(name: String, refType: String)
+    case binOp(op: String, left: ASTNode, right: ASTNode)
+    case unaryOp(op: String, operand: ASTNode)
+    case ternary(cond: ASTNode, then: ASTNode, `else`: ASTNode)
+    case member(obj: ASTNode, field: String)
+    case index(obj: ASTNode, index: ASTNode)
+    case assign(target: String, value: ASTNode)
+
+    public var isRuntimeExpr: Bool {
+        switch self {
+        case .stateRef, .runtimeRef, .binOp, .unaryOp, .ternary, .member, .index, .assign:
+            return true
+        default:
+            return false
+        }
+    }
+}
+
+public struct CallNode {
+    public let callee: String
+    public let args: [ASTNode]
+}
+
+public enum Statement {
+    case value(id: String, expr: ASTNode)
+    case state(id: String, initExpr: ASTNode)
+    case query(id: String, call: CallNode, expr: ASTNode, deps: [String]?)
+    case mutation(id: String, call: CallNode, expr: ASTNode)
+}
--- a/packages/swift/Sources/OpenUILang/Library.swift
+++ b/packages/swift/Sources/OpenUILang/Library.swift
@@ -1,0 +1,33 @@
+import Foundation
+
+public struct ComponentDef {
+    public let name: String
+    public let description: String
+    public let signature: String
+    
+    public init(name: String, description: String, signature: String) {
+        self.name = name
+        self.description = description
+        self.signature = signature
+    }
+}
+
+public protocol AnyComponentRenderer {
+    // In OpenUISwiftUI we will cast this to AnyView
+    func render(props: [String: Any], children: [Any]) -> Any
+}
+
+public class ComponentLibrary {
+    public private(set) var components: [String: ComponentDef] = [:]
+    public var root: String?
+    
+    public init() {}
+    
+    public func register(component: ComponentDef) {
+        components[component.name] = component
+    }
+    
+    public func setRoot(_ root: String) {
+        self.root = root
+    }
+}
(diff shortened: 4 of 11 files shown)


ankit-thesys:
First off, nice start. The package builds cleanly on Mac, the overall structure is sensible, and the prompt generator and basic SwiftUI rendering work. As a foundation, it's a reasonable first step.

There's one big problem, though: it doesn't actually speak OpenUI Lang. The parser was built against a made-up syntax that only looks like OpenUI Lang. They share the name = Component(...) shape and nothing else. I tested it by feeding it real examples straight from our own docs and tests — and every one came back empty, with no error reported. So a Swift app built on this wouldn't be compatible with any of our other runtimes (React, Vue, Svelte), which defeats the whole "one portable language everywhere" goal of the issue.

  The three things that make it a different language:
  - Argument style is backwards. Real OpenUI Lang uses positional args — Button("Submit", "submit:signup").
  This parser only accepts named args (Button(label: "...")) and silently throws positional ones away.
  - References don't work. Real OpenUI Lang splits UI across lines that reference each other (root = 
  Stack([header]), then header = ...). This parser drops those references, so anything nested disappears.
  - None of the dynamic stuff exists — variables, conditionals, queries. The types are defined but nothing
  actually parses them.

  A few smaller things worth flagging:
  - The tests don't run — the test file isn't wired into the package, so it ships with zero working tests.
  - Only 4 components are supported (fine for a first version), and button actions just print to the 
  console — there's no way for the app to actually react to them, which the issue specifically asked for.
 
  Bottom line: good scaffolding, wrong language. The fix is real work but well-scoped — make it positional,
  resolve references, and follow the existing grammar in packages/lang-core/src/parser/, which already
  defines exactly how this should behave. I'd treat this as "phase 1, needs another pass" rather than ready
  to merge.

amankoli09:
@ankit-thesys  i have made the change please have a look thanks

vishxrad (commented):
- **P0: Parser can hang on valid OpenUI syntax.** `parseArray` / `parseArguments` do not advance when an expression fails, so `{}`, `.member`, `@Builtin`, or other unsupported tokens inside arrays/args can spin forever. Valid spec examples like `Query("tool", {}, {rows: []})` are dangerous. See [Parser.swift#L149](https://github.com/thesysdev/openui/blob/5334a2529497ebc5ce54a9ccef784b575bbdb6b1/packages/swift/Sources/OpenUILang/Parser.swift#L149) and [Parser.swift#L164](https://github.com/thesysdev/openui/blob/5334a2529497ebc5ce54a9ccef784b575bbdb6b1/packages/swift/Sources/OpenUILang/Parser.swift#L164).

- **P0: Query/Mutation/state surfaces are fake.** The parser recognizes these statements, then returns `stateDeclarations: [:]`, `queryStatements: []`, and `mutationStatements: []` unconditionally. This makes live data, reactive state, and mutations unusable. See [Parser.swift#L42](https://github.com/thesysdev/openui/blob/5334a2529497ebc5ce54a9ccef784b575bbdb6b1/packages/swift/Sources/OpenUILang/Parser.swift#L42) and [Parser.swift#L75](https://github.com/thesysdev/openui/blob/5334a2529497ebc5ce54a9ccef784b575bbdb6b1/packages/swift/Sources/OpenUILang/Parser.swift#L75).

- **P0: Dynamic expressions are destroyed.** `$isLoading ? ... : ...` becomes the literal string `"$ternary"`, not an evaluable AST or rendered value. The demo’s own simulated stream will render placeholder text instead of the intended result. See [Parser.swift#L287](https://github.com/thesysdev/openui/blob/5334a2529497ebc5ce54a9ccef784b575bbdb6b1/packages/swift/Sources/OpenUILang/Parser.swift#L287) and the test that codifies this wrong behavior at [OpenUILangTests.swift#L58](https://github.com/thesysdev/openui/blob/5334a2529497ebc5ce54a9ccef784b575bbdb6b1/packages/swift/Tests/OpenUILangTests/OpenUILangTests.swift#L58).

- **P1: This is not OpenUI Lang v0.5 compatible.** It lacks objects, `null`, member access, binary/unary ops, `@` builtins, actions, escaped strings, negative numbers, and real Query/Mutation handling. The parser only supports a tiny toy subset. The unsupported surface starts around [Parser.swift#L96](https://github.com/thesysdev/openui/blob/5334a2529497ebc5ce54a9ccef784b575bbdb6b1/packages/swift/Sources/OpenUILang/Parser.swift#L96).

- **P1: Validation is essentially absent.** Unknown components, missing required props, null required props, and excess args are not reported in `meta.errors`. With a library present, unknown components can become empty `ElementNode`s. See [Parser.swift#L238](https://github.com/thesysdev/openui/blob/5334a2529497ebc5ce54a9ccef784b575bbdb6b1/packages/swift/Sources/OpenUILang/Parser.swift#L238).

- **P1: Streaming metadata is false.** `meta.incomplete` is always `false`, `partial` is always `false`, and there is no partial/auto-close behavior. The demo simulates token arrival, but the parser does not expose streaming state correctly. See [Parser.swift#L74](https://github.com/thesysdev/openui/blob/5334a2529497ebc5ce54a9ccef784b575bbdb6b1/packages/swift/Sources/OpenUILang/Parser.swift#L74) and [Parser.swift#L259](https://github.com/thesysdev/openui/blob/5334a2529497ebc5ce54a9ccef784b575bbdb6b1/packages/swift/Sources/OpenUILang/Parser.swift#L259).

- **P1: Reference cycles can recurse forever.** `a = b; b = a` has no visited-set protection and can stack overflow. See [Parser.swift#L227](https://github.com/thesysdev/openui/blob/5334a2529497ebc5ce54a9ccef784b575bbdb6b1/packages/swift/Sources/OpenUILang/Parser.swift#L227).

- **P2: Renderer is hardcoded, not library-driven.** `ComponentLibrary` cannot register renderers, and `OpenUIRenderer` only knows `VStack`, `HStack`, `Text`, and `Button`. That is not a useful native component integration story. See [Library.swift#L37](https://github.com/thesysdev/openui/blob/5334a2529497ebc5ce54a9ccef784b575bbdb6b1/packages/swift/Sources/OpenUILang/Library.swift#L37) and [OpenUIRenderer.swift#L26](https://github.com/thesysdev/openui/blob/5334a2529497ebc5ce54a9ccef784b575bbdb6b1/packages/swift/Sources/OpenUISwiftUI/OpenUIRenderer.swift#L26).

- **P2: The “chat” example is misleading.** It has no model call, no backend, no tool provider, no real stream, and no chat loop. It is a hardcoded text replay. See [App.swift#L88](https://github.com/thesysdev/openui/blob/5334a2529497ebc5ce54a9ccef784b575bbdb6b1/packages/swift/Sources/SwiftUIChat/App.swift#L88).

- **P3: Tests are far too weak.** Four tests pass, but they mostly prove the toy path works. They miss valid spec samples, invalid syntax, streaming partials, Query/Mutation extraction, objects, member access, builtins, cycles, and renderer behavior. [scratch.swift](https://github.com/thesysdev/openui/blob/5334a2529497ebc5ce54a9ccef784b575bbdb6b1/packages/swift/scratch.swift) also should not be committed.

**Verdict**

Not approved. It is a rough prototype with a Swift package shape, but the parser/runtime contract is not compatible with OpenUI Lang in the ways that matter. Passing `swift test` here is not meaningful because the tests mostly bless the broken behavior.

PR #605, opened 2026-06-05
fix(parser): preserve markdown fences and comments inside strings

### LINKED ISSUE #581: Markdown code blocks are not visible (also in https://www.openui.com/playground)
Example prompt:

```
give me js code example with `console.log("Hello World")` in markdown (use MarkDownRenderer instead of CodeBlock)
```

```
root = Stack([md])
md = MarkDownRenderer("```js\nconsole.log(\"Hello World\");\n```")
```

Output: nothing

Image

### LINKED ISSUE #596: Markdown text that includes `//` in multiline strings is not visible
Example:

```
root = MarkDownRenderer("
Hello // World
https://example.com/
")
```

Actual output:

```
Hello
https:
```

Expected output:

```
Hello // World
https://example.com/
```

Actual parsing output:

```
{
  "root": {
    "type": "element",
    "typeName": "MarkDownRenderer",
    "props": {
      "textMarkdown": "\nHello\nhttps:\n"
    },
    "partial": false,
    "hasDynamicProps": false,
    "statementId": "root"
  },
  "meta": {
    "incomplete": false,
    "unresolved": [],
    "orphaned": [],
    "statementCount": 1,
    "errors": []
  },
  "stateDeclarations": {},
  "queryStatements": [],
  "mutationStatements": []
}
```

shogun444:
## What

This PR fixes two parser preprocessing issues where markdown content inside string literals could be incorrectly modified before parsing.

## Changes

* Made `stripFences` string-aware so fenced code blocks inside string literals are preserved.
* Preserved string context across line boundaries in `stripComments`.
* Added regression tests covering both issues.

## Test Plan

* [ ] Not applicable (explain why)
* [x] Verified locally

```bash
pnpm --filter @openuidev/lang-core test
```

## Checklist

* [x] I linked a related issue, if applicable
* [ ] I updated docs/README when needed
* [x] I considered backwards compatibility

Closes #581

Closes #596

### DIFF
--- a/packages/lang-core/src/parser/parser.ts
+++ b/packages/lang-core/src/parser/parser.ts
@@ -253,8 +253,43 @@ export function stripFences(input: string): string {
   let i = 0;
 
   while (i < input.length) {
-    // Look for opening ```
-    const fenceStart = input.indexOf("```", i);
+    // Scan for opening ``` while tracking string context
+    let inStr: false | '"' | "'" = false;
+    let fenceStart = -1;
+    while (i < input.length) {
+      const c = input[i];
+      if (inStr) {
+        if (c === "\\" && i + 1 < input.length) {
+          i += 2; // skip escaped character
+          continue;
+        }
+        if (c === inStr) {
+          inStr = false;
+        }
+        i++;
+        continue;
+      }
+
+      // Not in string
+      if (c === '"' || c === "'") {
+        inStr = c;
+        i++;
+        continue;
+      }
+
+      if (
+        c === "`" &&
+        i + 1 < input.length &&
+        input[i + 1] === "`" &&
+        i + 2 < input.length &&
+        input[i + 2] === "`"
+      ) {
+        fenceStart = i;
+        break;
+      }
+      i++;
+    }
+
     if (fenceStart === -1) break;
 
     // Skip language tag until newline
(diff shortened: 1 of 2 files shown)


packages/lang-core/src/parser/parser.ts
```diff
@@ -269,23 +304,23 @@ export function stripFences(input: string): string {
...
         }
-        if (c === '"') inStr = false;
+        if (c === closeInStr) closeInStr = false;
         k++;
         continue;
       }
       // Not in string
-      if (c === '"') {
-        inStr = true;
+      if (c === '"' || c === "'") {
```
Aditya-thesys: Adding single quote tracking will cause issues with Apostrophes. 
LLMs many times generates at the start
`Here's the code: `
Which now will not strip the following fences.
Could you for now scope it to double quotes only.

packages/lang-core/src/parser/parser.ts
```diff
@@ -253,8 +253,43 @@ export function stripFences(input: string): string {
...
+      const c = input[i];
+      if (inStr) {
+        if (c === "\\" && i + 1 < input.length) {
+          i += 2; // skip escaped character
+          continue;
+        }
+        if (c === inStr) {
+          inStr = false;
+        }
+        i++;
```
Aditya-thesys: nit: the inStr and closeStr tracking code is almost identical, we can deduplicate them

Aditya-thesys (commented):
All looks good, could you fix those two small changes. Also maybe update the tests for single quote.

packages/lang-core/src/parser/parser.ts
shogun444: Nice catch!|
I'll do this right away.

packages/lang-core/src/parser/parser.ts
shogun444: yeah i will use a helper here.

shogun444:
Done!

- [x] Scoped fence tracking to double quotes only
- [x] Deduplicated the scanner logic with a helper
- [x] Added a regression test for apostrophes before fenced code blocks

Aditya-thesys (approved):

Aditya-thesys:
Thanks! for the fix, merging it.

PR #612, opened 2026-06-06
chore: centralize shared dependency versions with pnpm catalog

Shinyaigeek:
## Summary

Introduces a [pnpm catalog](https://pnpm.io/catalogs) in `pnpm-workspace.yaml` for dependencies that were already pinned to an **identical version across multiple packages**. Each version now lives in one place and is referenced from `package.json` via the `catalog:` protocol, so future bumps happen in a single spot.
**This is just a proposal, so if you don't think this is needed, please feel free to close this pull request**

Scoped intentionally to deps that already shared one version spec — so this is a **pure refactor with no resolved-version changes**. I think some devDependencies(like a vitest)'s version also should be unified by this feature, but some of these have different version, so I fix this later if this approach suites this project.

## Catalog entries

```yaml
catalog:
  jsdom: "^26.1.0"
  zod: "^3.25.0 || ^4.0.0"
  zustand: "^4.5.5"
```

| Dependency | Spec | Packages updated to `catalog:` |
| --- | --- | --- |
| `zod` | `^3.25.0 \|\| ^4.0.0` | lang-core, react-lang, react-ui, react-email, vue-lang, svelte-lang |
| `zustand` | `^4.5.5` | react-headless, react-ui |
| `jsdom` | `^26.1.0` | svelte-lang, vue-lang |

## Notes
Deps with differing versions across packages (e.g. `vitest` v3 vs v4, `react`, `vite`) were deliberately left out of this PR to avoid bundling version unifications into a refactor. They can be cataloged in a follow-up if desired.

### DIFF
--- a/packages/lang-core/package.json
+++ b/packages/lang-core/package.json
@@ -41,7 +41,7 @@
   },
   "peerDependencies": {
     "@modelcontextprotocol/sdk": ">=1.0.0",
-    "zod": "^3.25.0 || ^4.0.0"
+    "zod": "catalog:"
   },
   "peerDependenciesMeta": {
     "@modelcontextprotocol/sdk": {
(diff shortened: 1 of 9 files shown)


abhithesys (approved):
LGTM

PR #628, opened 2026-06-10
fix(build): fix tsdown external dependency regex on Windows (#626)

### LINKED ISSUE #626: Fix tsdown external dependency regex on Windows
Running `pnpm install` on Windows currently fails while building `@openuidev/browser-bundle` with errors like:

```txt
Could not resolve "./openuiLibrary.tsx"
Could not resolve "./library.ts"
Could not resolve "./Renderer.tsx"
```

The affected `tsdown` configurations use:

```ts
deps: {
  neverBundle: [/^[^./]/],
}
```

This regex is intended to match bare package imports. However, on Windows, absolute paths begin with a drive letter, for example `C:\repo\...`. These paths also match the regex, so local files are incorrectly treated as external dependencies.

As a result, references to local TypeScript files are left in the generated output, causing `@openuidev/browser-bundle` to fail during its `prepare` build.

## Proposed fix

Replace the regex with a cross-platform `neverBundle` filter that excludes:

* relative imports
* POSIX absolute paths
* Windows absolute paths

```ts
const neverBundle = (id: string) =>
  !id.startsWith(".") &&
  !id.startsWith("/") &&
  !/^[A-Za-z]:[\\/]/.test(id);
```

Then use this function in the affected `tsdown` configurations.

## Acceptance criteria

* `pnpm install` completes successfully on Windows.
* Local modules are bundled instead of being treated as external.
* `@openuidev/browser-bundle` builds successfully.
* Existing Linux and macOS builds continue to pass.

This same Windows absolute-path issue was independently fixed in LangChain’s `deepagentsjs` [PR #489](https://github.com/langchain-ai/deepagentsjs/pull/489).

shogun444:
## What

Fixes #626.

The existing `neverBundle` regex (`/^[^./]/`) treats Windows drive-letter paths as external dependencies. During Windows builds this leaves local TypeScript imports unresolved in generated output, which later causes `@openuidev/browser-bundle` to fail with errors such as:

```txt
Could not resolve "./library.ts"
Could not resolve "./Renderer.tsx"
```

This change updates the regex so Windows absolute paths are handled correctly while preserving the existing behavior for package imports.

## Changes

Updated the `neverBundle` configuration in the affected `tsdown` configs to avoid matching Windows absolute paths.

## Test Plan

Tested on Windows 11.

`pnpm install` now completes successfully and `@openuidev/browser-bundle` builds without the unresolved import errors. I also verified that `pnpm --filter @openuidev/docs dev` starts successfully after the change.

### DIFF
--- a/packages/lang-core/tsdown.config.ts
+++ b/packages/lang-core/tsdown.config.ts
@@ -9,6 +9,6 @@ export default defineConfig({
   outDir: "dist",
   clean: true,
   deps: {
-    neverBundle: [/^[^./]/],
+    neverBundle: [/^(?![./]|[A-Za-z]:[/\\])/],
   },
 });
--- a/packages/react-email/tsdown.config.ts
+++ b/packages/react-email/tsdown.config.ts
@@ -9,6 +9,6 @@ export default defineConfig({
   outDir: "dist",
   clean: true,
   deps: {
-    neverBundle: [/^[^./]/],
+    neverBundle: [/^(?![./]|[A-Za-z]:[/\\])/],
   },
 });
--- a/packages/react-headless/tsdown.config.ts
+++ b/packages/react-headless/tsdown.config.ts
@@ -9,6 +9,6 @@ export default defineConfig({
   outDir: "dist",
   clean: true,
   deps: {
-    neverBundle: [/^[^./]/],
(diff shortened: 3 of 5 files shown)


vishxrad (approved):
LGTM.

PR #634, opened 2026-06-11
chore: centralize eslint toolchain versions with pnpm catalog

Shinyaigeek:
## What

Move the eslint toolchain version specs into the pnpm `catalog:` and reference them from the consuming manifests, so each version is declared in one place. Follows the pattern established in #612.

## Why

The eslint toolchain was declared with **identical** version specs in two manifests — the root `package.json` and `packages/react-ui/package.json`. Centralizing them in the catalog removes the duplication and keeps the two in lockstep going forward.

## Cataloged

No declared-version changes — root and react-ui already shared one spec each:

| dep | spec |
| --- | --- |
| `@typescript-eslint/eslint-plugin` | `^8.56.1` |
| `eslint` | `^9.0.0` |
| `eslint-config-prettier` | `^10.1.8` |
| `eslint-plugin-prettier` | `^5.5.5` |
| `eslint-plugin-react-hooks` | `^7.0.1` |
| `eslint-plugin-react-refresh` | `^0.5.2` |
| `eslint-plugin-storybook` | `^10.2.14` |
| `eslint-plugin-unused-imports` | `^4.4.1` |

## Scope

`examples/` and `docs/` use a looser `eslint: ^9` plus `eslint-config-next` (which no workspace package uses), so folding them in would change resolved specs rather than just centralizing them. Left out to match #612's "no resolved-version changes" principle. The other `packages/*` don't declare eslint themselves (they rely on hoisting), so there was nothing to convert.

## Lockfile churn

The diff is large because `pnpm install` deduped a stale transitive `@typescript-eslint` `8.56.1` sub-tree onto `8.59.4`. The top-level `@typescript-eslint/eslint-plugin` already resolved to `8.59.4` before this change, so no top-level resolution moved.

## Verification

- `pnpm install` clean (only pre-existing, unrelated peer warnings)
- `pnpm --filter @openuidev/react-ui run lint:check` passes — eslint `v9.29.0` resolves from the catalog, plugins load, **0 errors** (pre-existing `react-hooks/exhaustive-deps` warnings only)


vishxrad (approved):
LGTM. Thank you for your contribution :)

PR #624, opened 2026-06-09
chore: centralize react/react-dom/@types/react in pnpm catalog

Shinyaigeek:
## Summary

Follow-up to the pnpm catalog work (#612). Centralizes the React-related dependencies through the `catalog:` protocol so their versions live in one place.

The `react` / `react-dom` peer ranges were already aligned across the library packages but declared inline in each `package.json`, and `@types/react` was split across three equivalent-but-different ranges.

## Changes

Added to the `catalog` in `pnpm-workspace.yaml`:

| dep | catalog version |
| --- | --- |
| `react` | `^18.3.1 \|\| ^19.0.0` |
| `react-dom` | `^18.0.0 \|\| ^19.0.0` |
| `@types/react` | `>=19.0.0` |

Replaced the inline versions with `catalog:` in:

- `react-lang` — `react` (peer), `@types/react` (`^19.0.0` → catalog)
- `react-ui` — `react`, `react-dom` (peer), `@types/react` (`>=19.0.0` → catalog)
- `react-headless` — `react` (peer), `@types/react` (`>=19.0.0` → catalog)
- `react-email` — `react`, `react-dom` (peer), `@types/react` (`^19` → catalog)

`@types/react` is unified to `>=19.0.0` (matching `react-ui` / `react-headless`).

### Intentionally left as-is

- **`browser-bundle`** `react` / `react-dom` are runtime `dependencies` pinned to `^19.0.0` (the bundled version), not the wider peer range — they shouldn't share the peer catalog entry.
- **`@types/react-dom`** is used in a single package, so there's no version to centralize.
- **`openui-cli` chat template** is outside the workspace (`!**/src/templates/**`), so `catalog:` would not resolve there.

## Verification

- `pnpm install` resolves cleanly (remaining peer warnings are pre-existing and unrelated)
- `@openuidev/react-headless` tests: 70 passed / 70
- `react-lang` / `react-ui` have no test files (`--passWithNoTests`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #645, opened 2026-06-11
chore: centralize typescript version with pnpm catalog

Shinyaigeek:
## What

`typescript` was declared as a devDependency with divergent specs across the library packages and the monorepo root. Move the version into the pnpm catalog and reference it via the `catalog:` protocol so the toolchain version lives in one place — mirroring #612 / #624 / #634.

Cataloged spec: **`^5.9.3`** (the root's existing spec).

| manifest | before | after |
|---|---|---|
| `package.json` (root) | `^5.9.3` | `catalog:` |
| `packages/react-email` | `^5` | `catalog:` |
| `packages/svelte-lang` | `^5.0.0` | `catalog:` |
| `packages/vue-lang` | `^5.0.0` | `catalog:` |

## No resolved-version change

Every importer already resolved to `typescript@5.9.3`, so this only tightens the declared specs onto the single catalog entry — the lockfile diff is just the four specifiers flipping to `catalog:` plus the new catalog entry. `examples/` and `docs/` keep their inline `^5` / `~5.9.2` specs, matching #612's library-only scope.

## Drive-by fix

`fix(svelte-lang)`: the test `src/__tests__/library.test.ts` used dot access on `lib.components` (a string-index-signature type), which `noPropertyAccessFromIndexSignature` (set in the root tsconfig) rejects. Switched to bracket access, matching the parallel `vue-lang` test. This was already failing `svelte-check` on `main`; surfaced while verifying this change.

## Verification

- `react-email`, `svelte-lang`, `vue-lang`: **build + typecheck/check pass** with `typescript@5.9.3` resolved from the catalog.
- `svelte-lang` `vitest`: 30/30 pass.
- `pnpm install --frozen-lockfile`: clean (lockfile consistent with manifests).

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "prettier-plugin-organize-imports": "^3.2.4",
     "publint": "^0.3.18",
     "tsdown": "^0.21.7",
-    "typescript": "^5.9.3"
+    "typescript": "catalog:"
   },
   "keywords": [
     "openui",
--- a/packages/react-email/package.json
+++ b/packages/react-email/package.json
@@ -54,6 +54,6 @@
   },
   "devDependencies": {
     "@types/react": "catalog:",
-    "typescript": "^5"
+    "typescript": "catalog:"
   }
 }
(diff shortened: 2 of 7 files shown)


ankit-thesys (approved):
Looks good to me. 

Thank you for your contribution. :)

PR #651, opened 2026-06-13
fix(charts): prevent TypeError crash when streaming nullish data

### LINKED ISSUE #355: Charts component crashes with TypeError during streaming: null data passed to Recharts

serhiizghama:
## Problem

Closes #355.

Every public chart component (`BarChart`, `BarChartCondensed`, `LineChart`, `AreaChart`, `PieChart`, `RadarChart`, `ScatterChart`, …) reads its `data` prop directly — `getDataKeys(data, …)`, `data.length`, `[...data].sort()` — assuming `data` is always an array.

While generative UI is streaming, `data` can momentarily be `null`/`undefined` before the full payload has been parsed. Passing that into the chart internals throws:

```
TypeError: Cannot read properties of null (reading '0')   // getDataKeys
TypeError: Cannot read properties of null (reading 'length')
TypeError: data is not iterable                            // [...data]
```

This crashes the whole render mid-stream.

## Solution

Add a small shared helper `normalizeChartData()` in `Charts/utils/dataUtils.ts` that coerces `data` to an array, and call it once at the top of each chart component so the rest of the component always works with a real array (an empty chart instead of a crash).

Two details that make it safe:

- **Stable empty reference** — when `data` is nullish it falls back to a single shared `EMPTY_CHART_DATA` constant, so normalization never creates a new `[]` on every render (which would change identity and defeat the existing `useMemo`/`React.memo` optimizations). When `data` is a valid array it is returned by reference, unchanged.
- `getDataKeys()` is also made null-safe (`data?.[0]`) as defense-in-depth, since it is the shared first access point.

The `genui-lib` schema layer already guards its own path via `hasAllProps`/`buildChartData`; this PR hardens the **public exported chart components** themselves, which are part of the package surface and can be rendered directly by consumers (and during streaming).

This matches the approach agreed on in the issue thread (`Array.isArray(data) ? data : []`, applied consistently across all chart types).

## Testing

- `pnpm --filter @openuidev/react-ui typecheck` — passes
- `pnpm --filter @openuidev/react-ui ci` (`lint:check` + `format:check`) — passes
- Verified `normalizeChartData(null)`/`(undefined)` return the same stable empty array, valid arrays pass through by reference, and `getDataKeys(null)` returns `[]` instead of throwing.

### DIFF
--- a/packages/react-ui/src/components/Charts/AreaChart/AreaChart.tsx
+++ b/packages/react-ui/src/components/Charts/AreaChart/AreaChart.tsx
@@ -35,6 +35,7 @@ import {
   getColorForDataKey,
   getDataKeys,
   getLegendItems,
+  normalizeChartData,
 } from "../utils/dataUtils";
 import { AreaChartData, AreaChartVariant } from "./types";
 
@@ -83,6 +84,8 @@ const AreaChartComponent = <T extends AreaChartData>({
   height,
   width,
 }: AreaChartProps<T>) => {
+  // Guard against nullish `data` while generative UI is streaming.
+  data = normalizeChartData(data);
   const printContext = usePrintContext();
   isAnimationActive = printContext ? false : isAnimationActive;
 
--- a/packages/react-ui/src/components/Charts/AreaChartCondensed/AreaChartCondensed.tsx
+++ b/packages/react-ui/src/components/Charts/AreaChartCondensed/AreaChartCondensed.tsx
@@ -31,6 +31,7 @@ import {
   getColorForDataKey,
   getDataKeys,
   getLegendItems,
+  normalizeChartData,
 } from "../utils/dataUtils";
 import { PaletteName, useChartPalette } from "../utils/PalletUtils";
 
@@ -79,6 +80,8 @@ const AreaChartCondensedComponent = <T extends AreaChartData>({
   height = CHART_HEIGHT,
   width,
 }: AreaChartCondensedProps<T>) => {
+  // Guard against nullish `data` while generative UI is streaming.
+  data = normalizeChartData(data);
   const printContext = usePrintContext();
   isAnimationActive = printContext ? false : isAnimationActive;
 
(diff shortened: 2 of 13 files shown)


vishxrad:
Thanks for the PR. We’re going to address this at the genui-lib streaming/parser layer rather than in each chart component. Please keep this PR open for now while we implement and verify that approach. Once the replacement is confirmed, we’ll close this PR as superseded. We’ll track the work in #355.

PR #668, opened 2026-06-21
fix(react-ui): avoid nested button in foldable section trigger

rabisg:
## What

The foldable section trigger rendered an `IconButton` (a `<button>`) inside Radix's `AccordionTrigger` (also a `<button>`). Nested `<button>` elements are invalid HTML and caused a React hydration error:

> In HTML, `<button>` cannot be a descendant of `<button>`. This will cause a hydration error.

## Changes

- Made `IconButton` polymorphic via Radix `Slot` + `Slottable` with a new `asChild` prop, so it can render as a caller-provided element instead of a `<button>` while preserving the inner `.openui-icon-button-icon` wrapper (required by the size CSS descendant selectors).
- Updated `FoldableSectionTrigger` to render the decorative chevron with `asChild` + an `aria-hidden` `<span>` host, eliminating the nested `<button>`.
- Added `@radix-ui/react-slot` as a direct dependency of `@openuidev/react-ui` (with the corresponding minimal `pnpm-lock.yaml` importer entry).

## Test Plan

- [x] Verified locally
  - `pnpm --filter @openuidev/react-ui typecheck` passes.
  - `pnpm install --frozen-lockfile` passes (lockfile consistent).
  - Default (non-`asChild`) `IconButton` usage is unchanged; only the foldable trigger renders the chevron as a span, so the nested-button hydration error no longer occurs.

## Checklist

- [ ] I updated docs/README when needed

### DIFF
--- a/packages/react-ui/package.json
+++ b/packages/react-ui/package.json
@@ -112,6 +112,7 @@
     "@radix-ui/react-select": "^2.1.5",
     "@radix-ui/react-separator": "^1.1.7",
     "@radix-ui/react-slider": "^1.2.2",
+    "@radix-ui/react-slot": "^1.2.4",
     "@radix-ui/react-switch": "^1.1.2",
     "@radix-ui/react-tabs": "^1.1.2",
     "@radix-ui/react-toggle-group": "^1.1.2",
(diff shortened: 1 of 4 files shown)


i-subham23 (approved):
L.G.

PR #710, opened 2026-07-02
fix(#701): persist thread messages before switching to new chat

### LINKED ISSUE #701: New Chat button clears current session instead of preserving it as a thread
## Describe the bug
Clicking "New Chat" button causes the current chat session to disappear entirely. The conversation is not saved as a previous thread and becomes inaccessible.

## To Reproduce
1. Fresh scaffold: `npx @openuidev/cli@latest create`
2. Set up `.env` with Ollama local endpoint
3. Send a few messages in the chat
4. Click the "New Chat" button in the UI
5. The previous session disappears — no thread in sidebar, no way to get back

## Expected behavior
Previous session should be saved as a thread in the sidebar, allowing users to switch between multiple conversations like a standard chat app.

## Screenshots

**Before clicking New Chat** : Normal chat with AI-generated UI components visible on left

Before

**After clicking New Chat** : Same session  components on left panel disappeared after clicking New Chat

After

## Desktop
 - OS: Windows 11
 - Browser: (tera browser)
 - Version: (browser version)

## Additional context
Confirmed on both fresh CLI scaffold and upgraded project. Likely related to default in-memory storage in ChatProvider not persisting threads on session switch.

shogun444:
Closes #701

## What

This PR addresses the message persistence issue when switching to a new chat by introducing an optional `cacheMessages` method to `ThreadStorage`.

The default in-memory storage implements this method by caching the current messages before the local state is cleared, allowing completed conversations to be restored when switching back to the thread.

## Changes

- Add optional `cacheMessages(threadId, messages)` to `ThreadStorage`
- Implement `cacheMessages` in the default in-memory storage
- Cache current messages before `switchToNewThread()` clears the local state
- Persist messages after successful LLM streaming completes
- Add unit tests covering both persistence paths

> **Note:** While validating this change, I noticed another reproducible issue when switching threads during an active streaming response. I've documented that separately in the issue discussion since it appears to be a different edge case.

## Test Plan

- [x] Verified locally
- [x] Added unit tests covering both persistence paths

## Checklist

- [x] I linked a related issue
- [ ] I updated docs/README when needed (not required)

### DIFF
--- a/packages/react-headless/src/store/createChatStore.ts
+++ b/packages/react-headless/src/store/createChatStore.ts
@@ -74,6 +74,10 @@ export const createChatStore = (config: CreateChatStoreConfig) => {
 
       switchToNewThread: () => {
         get().cancelMessage();
+        const { selectedThreadId, messages } = get();
+        if (selectedThreadId && messages.length > 0) {
+          threadStorage.cacheMessages?.(selectedThreadId, messages).catch(() => {});
+        }
         set({
           selectedThreadId: null,
           messages: [],
@@ -201,6 +205,9 @@ export const createChatStore = (config: CreateChatStoreConfig) => {
               }),
             adapter: llm.streamProtocol,
           });
+
+          // Persist messages after successful streaming so they survive thread switches.
+          await threadStorage.cacheMessages?.(threadId, get().messages);
         } catch (e) {
           if (!abortController.signal.aborted) {
             set({ threadError: e instanceof Error ? e : new Error(String(e)) });
--- a/packages/react-headless/src/adapters/_defaultStorage.ts
+++ b/packages/react-headless/src/adapters/_defaultStorage.ts
@@ -35,6 +35,9 @@ export function createDefaultInMemoryStorage(): ChatStorage {
       async getMessages(threadId: string) {
         return messagesByThread.get(threadId) ?? [];
       },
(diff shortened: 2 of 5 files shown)


shogun444:
While testing this a bit more, I noticed another case that may be related.

If I send a message and the assistant starts streaming a response (for example, generating a component), then quickly switch to another chat or click **New Chat** before the response finishes, and later come back to the original thread, the message I originally sent is sometimes gone as well.

From what I've seen:
- Start a new conversation.
- Send a prompt that produces a longer streamed response.
- While it's still streaming, switch to another thread or click **New Chat**.
- Return to the original thread.

Actual:
- In some cases, the user message itself disappears from the conversation.

I'm not sure if this is the same underlying issue or a separate race condition, but I was able to reproduce it multiple times while testing.

packages/react-headless/src/store/createChatStore.ts
```diff
@@ -74,6 +74,10 @@ export const createChatStore = (config: CreateChatStoreConfig) => {
 
       switchToNewThread: () => {
         get().cancelMessage();
+        const { selectedThreadId, messages } = get();
+        if (selectedThreadId && messages.length > 0) {
+          threadStorage.cacheMessages?.(selectedThreadId, messages).catch(() => {});
```
abhithesys, line 79: why is this needed?

PR #715, opened 2026-07-04
fix(react-ui): stop false "unknown key" warnings for valid *ChartPalette theme keys

### LINKED ISSUE #714: ThemeProvider falsely warns "unknown key — it will be ignored" for valid *ChartPalette theme keys
## Situation

Theming charts the typed, documented way compiles fine but spams dev-console warnings on every boot. Reported by a downstream consumer on `@openuidev/react-ui@0.12.1`, confirmed on `main`:

```tsx
const t = createTheme({ defaultChartPalette: ["#57C8D6", "#FF6A2B"] }); // compiles
<ThemeProvider mode="dark" darkTheme={t}>…</ThemeProvider>
```

```
[OpenUI] Unknown theme key "defaultChartPalette".
[OpenUI] darkTheme contains unknown key "defaultChartPalette". It will be ignored. Use createTheme() for typo detection with suggestions.
```

Every `*ChartPalette` key is warned twice — once by `createTheme()`, once by the ThemeProvider prop validator. The consumer's workaround was to abandon theme-level palettes and pass `customPalette` on every chart.

## Task

Establish whether the palettes are genuinely ignored (functional bug) or the warning is a false positive (types-vs-runtime drift), then eliminate the false warnings without losing typo detection, and prevent the drift from recurring.

## Action (investigation findings)

**Root cause — type and runtime allow-list drifted:**

- `packages/react-ui/src/components/ThemeProvider/types.ts:8-17` — `ChartColorPalette` declares **eight** optional `string[]` keys (`defaultChartPalette`, `barChartPalette`, `lineChartPalette`, `areaChartPalette`, `pieChartPalette`, `radarChartPalette`, `radialChartPalette`, `horizontalBarChartPalette`); `Theme` includes them via `ColorTheme extends ChartColorPalette` → TypeScript accepts them.
- `ThemeProvider.tsx:112` and `utils.ts:70` both derive their allow-list from `Object.keys(defaultLightTheme)`, and `defaultTheme.ts:487` defines none of the palette keys → every palette key trips "unknown key" in both validators.

**Runtime verification (server-rendered probe through real `ThemeProvider` → `useTheme()` → `useChartPalette`):**

- The merge spreads user overrides wholesale (`{ ...themes.dark, ...overrides }`, `ThemeProvider.tsx:208-211`), so palettes **do** reach context and charts **do** apply them (`Charts/utils/PalletUtils.ts:185`); unset chart types fall back to `defaultChartPalette` correctly.
- Conclusion: **"It will be ignored" is false.** Dev-only misleading noise, not a functional regression — but it drives consumers off a working API.

## Result (definition of done)

- [ ] `CHART_PALETTE_KEYS` constant type-guarded with `satisfies Record<keyof ChartColorPalette, true>` — drift becomes a compile error
- [ ] Both validators consume one shared allow-list (`Object.keys(defaultLightTheme)` + palette keys)
- [ ] Zero warnings for valid palette keys; typo suggestions still work for real typos
- [ ] Regression tests: no false warning, palettes flow to `useTheme()`/`useChartPalette`, allow-list covers every declared palette key

ankit-thesys:
Fixes #714

**Superseded by #721** (branch renamed).

## Summary

- Theming charts the typed way — `createTheme({ defaultChartPalette: [...] })` on `lightTheme`/`darkTheme` — compiled fine but warned `unknown key … It will be ignored` twice per key on every dev boot. **The message was false**: the theme merge spreads user overrides wholesale, so palettes were reaching `useTheme()` and charts were applying them all along. Root cause: both validators derived their allow-list from `Object.keys(defaultLightTheme)`, which defines none of the eight `*ChartPalette` keys declared on the `Theme` type (types-vs-runtime drift).
- Fix: one shared allow-list (`KNOWN_THEME_KEYS` = default theme keys + `CHART_PALETTE_KEYS`) consumed by both `createTheme()` and the ThemeProvider prop validator. `CHART_PALETTE_KEYS` is guarded by `satisfies Record<keyof ChartColorPalette, true>`, so adding/removing a palette key on the type without updating the allow-list is now a **compile error**. Deliberately no palette defaults in `defaultTheme.ts` — a default would shadow the per-chart `theme` prop fallback in `useChartPalette`, and `themeToCssVars`'s string-only filter means palettes never belonged in CSS vars anyway.
- Added regression tests (`ThemeProvider/__tests__/ThemeProvider.test.tsx`) + `tsconfig.test.json` (required by the ESLint override for `__tests__` files, mirroring react-headless). Updated the ThemeProvider `AGENTS.md` with the palette-key exception to the "adding a new token" recipe.

## Investigation findings (verified against `main`)

| Claim | Verdict |
|---|---|
| Type declares palette keys, validators reject them | Confirmed — but it's **8** keys (`default/bar/line/area/pie/radar/radial/horizontalBar`), not 5 as first reported |
| Both validators use `Object.keys(defaultLightTheme)` | Confirmed (`ThemeProvider.tsx:112`, `utils.ts:70`); consumers using `createTheme` got warned **twice** per key |
| "It will be ignored" | **False** — runtime probe through real `ThemeProvider → useTheme() → useChartPalette` showed palettes survive the merge (`{ ...themes.dark, ...overrides }`) and drive chart colors, incl. fallback to `defaultChartPalette` for unset chart types |
| Severity | Dev-only misleading noise — but it drove a downstream consumer to abandon theme-level palettes for per-chart `customPalette` workarounds |

## Is the problem gone?

Yes — re-ran the exact pre-fix reproduction probe against this branch:

| | Before (main) | After (this PR) |
|---|---|---|
| Warnings on mount | 6 (2 per palette key) | **0** |
| `theme.barChartPalette` in context | present | present (unchanged) |
| Chart colors from theme palette | applied | applied (unchanged) |
| Typo detection (`defaultChartPalete`) | warns | still warns, now suggests `defaultChartPalette` |

## Test plan

- [x] `pnpm --filter @openuidev/react-ui run test` — 13/13 pass (7 new: allow-list covers every declared palette key + every default-theme key; `createTheme` silent for all 8 palette keys; ThemeProvider props silent for palette keys; genuine unknown keys still warn with suggestion; palettes flow theme → `useTheme()` → `useChartPalette` incl. `defaultChartPalette` fallback)
- [x] `pnpm --filter @openuidev/react-ui run typecheck` — clean
- [x] `pnpm --filter @openuidev/react-ui run lint:check` — 0 errors (pre-existing warnings only)
- [x] `pnpm --filter @openuidev/react-ui run format:check` — clean
- [x] Pre-fix repro probe re-run on branch: zero warnings, palettes still applied (table above)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/react-ui/src/components/ThemeProvider/AGENTS.md
+++ b/packages/react-ui/src/components/ThemeProvider/AGENTS.md
@@ -2,14 +2,14 @@
 
 ## File Map
 
-| File                | Purpose                                                                                                                                                    |
-| ------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `ThemeProvider.tsx` | React component, `ThemeContext`, `InternalContext` (nesting detection), `useTheme` hook. Injects `--openui-*` vars into `<head>` via `useInsertionEffect`. |
-| `types.ts`          | TypeScript interfaces: `Theme`, `ColorTheme`, `LayoutTheme`, `TypographyTheme`, `EffectTheme`, `ChartColorPalette`.                                        |
-| `defaultTheme.ts`   | Builds `defaultLightTheme` / `defaultDarkTheme` (frozen) from the swatch system. Contains `createColorTheme()` and all layout/typography/shadow defaults.  |
-| `swatches.ts`       | 18 oklch color families x 14 shades. Exports `swatch()`, `withAlpha()`, `swatchToken()`, `swatchTokens`.                                                   |
-| `utils.ts`          | Exports `camelToKebab`, `themeToCssVars`, `createTheme` (dev-mode typo detection with Levenshtein distance suggestions).                                   |
-| `index.ts`          | Barrel re-exports from all the above files.                                                                                                                |
+| File                | Purpose                                                                                                                                                                                               |
+| ------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `ThemeProvider.tsx` | React component, `ThemeContext`, `InternalContext` (nesting detection), `useTheme` hook. Injects `--openui-*` vars into `<head>` via `useInsertionEffect`.                                            |
+| `types.ts`          | TypeScript interfaces: `Theme`, `ColorTheme`, `LayoutTheme`, `TypographyTheme`, `EffectTheme`, `ChartColorPalette`.                                                                                   |
+| `defaultTheme.ts`   | Builds `defaultLightTheme` / `defaultDarkTheme` (frozen) from the swatch system. Contains `createColorTheme()` and all layout/typography/shadow defaults.                                             |
+| `swatches.ts`       | 18 oklch color families x 14 shades. Exports `swatch()`, `withAlpha()`, `swatchToken()`, `swatchTokens`.                                                                                              |
+| `utils.ts`          | Exports `camelToKebab`, `themeToCssVars`, `createTheme` (dev-mode typo detection with Levenshtein distance suggestions), and `KNOWN_THEME_KEYS` / `CHART_PALETTE_KEYS` (shared validator allow-list). |
+| `index.ts`          | Barrel re-exports from all the above files.                                                                                                                                                           |
 
 ## Architecture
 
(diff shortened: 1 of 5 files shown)


ankit-thesys:
Superseded by #721 — same change, branch renamed to drop an internal ticket reference from the branch name.

PR #721, opened 2026-07-06
fix(react-ui): stop false "unknown key" warnings for valid *ChartPalette theme keys

### LINKED ISSUE #714: ThemeProvider falsely warns "unknown key — it will be ignored" for valid *ChartPalette theme keys

ankit-thesys:
Fixes #714

## Background

Theming charts the typed way — `createTheme({ defaultChartPalette: [...] })` passed to `ThemeProvider`'s `lightTheme`/`darkTheme` — compiles cleanly but printed dev-console warnings on every boot, twice per palette key:

```
[OpenUI] Unknown theme key "defaultChartPalette".
[OpenUI] darkTheme contains unknown key "defaultChartPalette". It will be ignored. Use createTheme() for typo detection with suggestions.
```

The `Theme` type declares eight `*ChartPalette` keys (`default`/`bar`/`line`/`area`/`pie`/`radar`/`radial`/`horizontalBar`), but both validators derived their allow-list from `Object.keys(defaultLightTheme)` (`ThemeProvider.tsx:112`, `utils.ts:70`), and the default themes define none of those keys — so every valid palette key tripped "unknown key". A runtime probe through the real `ThemeProvider → useTheme() → useChartPalette` chain showed the warning text was wrong: the theme merge spreads user overrides wholesale, so the palettes were never ignored — charts applied them all along, including the fallback to `defaultChartPalette` for unset chart types. Dev-only noise, but convincing enough that a downstream consumer abandoned theme-level palettes for per-chart `customPalette` workarounds.

## Goal

Make the typed chart-palette API usable without false warnings, keep typo detection for genuinely unknown keys, and ensure the type-vs-runtime drift that caused this cannot silently recur.

## What this PR does

- Introduces a single shared allow-list: `KNOWN_THEME_KEYS` = `Object.keys(defaultLightTheme)` + `CHART_PALETTE_KEYS`, consumed by both `createTheme()` and the ThemeProvider prop validator (previously two independent copies).
- Derives the `ChartColorPalette` type from a single `CHART_PALETTE_KEYS` `as const` array in `types.ts` — a palette key is written in exactly one place, so type/runtime drift is impossible by construction (no guarded mirror to maintain).
- Deliberately adds no palette defaults to `defaultTheme.ts`: a default palette would shadow the per-chart `theme` prop fallback in `useChartPalette`, and `themeToCssVars` filters to string values, so array palettes never belonged in CSS vars anyway.
- Adds regression tests (`ThemeProvider/__tests__/ThemeProvider.test.tsx`) plus `tsconfig.test.json` (required by the ESLint override for `__tests__` files, mirroring react-headless), and updates the ThemeProvider `AGENTS.md` with the palette-key exception to the "adding a new token" recipe.

## Outcome

Re-ran the exact pre-fix reproduction probe against this branch:

| | Before (main) | After (this PR) |
|---|---|---|
| Warnings on mount | 6 (2 per palette key) | **0** |
| `theme.barChartPalette` in context | present | present (unchanged) |
| Chart colors from theme palette | applied | applied (unchanged) |
| Typo detection (`defaultChartPalete`) | warns | still warns, now suggests `defaultChartPalette` |

Verification:

- [x] `pnpm --filter @openuidev/react-ui run test` — 13/13 pass (7 new: allow-list covers every declared palette key + every default-theme key; `createTheme` silent for all 8 palette keys; ThemeProvider props silent for palette keys; genuine unknown keys still warn with suggestion; palettes flow theme → `useTheme()` → `useChartPalette` incl. `defaultChartPalette` fallback)
- [x] `pnpm --filter @openuidev/react-ui run typecheck` — clean
- [x] `pnpm --filter @openuidev/react-ui run lint:check` — 0 errors (pre-existing warnings only)
- [x] `pnpm --filter @openuidev/react-ui run format:check` — clean

### DIFF
--- a/packages/react-ui/tsconfig.test.json
+++ b/packages/react-ui/tsconfig.test.json
@@ -1,0 +1,9 @@
+{
+  "$schema": "https://json.schemastore.org/tsconfig",
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "noEmit": true
+  },
+  "include": ["src/**/*"],
+  "exclude": ["dist", "node_modules"]
+}
--- a/packages/react-ui/src/components/ThemeProvider/AGENTS.md
+++ b/packages/react-ui/src/components/ThemeProvider/AGENTS.md
@@ -2,14 +2,14 @@
 
 ## File Map
 
-| File                | Purpose                                                                                                                                                                                               |
-| ------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `ThemeProvider.tsx` | React component, `ThemeContext`, `InternalContext` (nesting detection), `useTheme` hook. Injects `--openui-*` vars into `<head>` via `useInsertionEffect`.                                            |
-| `types.ts`          | TypeScript interfaces: `Theme`, `ColorTheme`, `LayoutTheme`, `TypographyTheme`, `EffectTheme`, `ChartColorPalette`.                                                                                   |
-| `defaultTheme.ts`   | Builds `defaultLightTheme` / `defaultDarkTheme` (frozen) from the swatch system. Contains `createColorTheme()` and all layout/typography/shadow defaults.                                             |
-| `swatches.ts`       | 18 oklch color families x 14 shades. Exports `swatch()`, `withAlpha()`, `swatchToken()`, `swatchTokens`.                                                                                              |
-| `utils.ts`          | Exports `camelToKebab`, `themeToCssVars`, `createTheme` (dev-mode typo detection with Levenshtein distance suggestions), and `KNOWN_THEME_KEYS` / `CHART_PALETTE_KEYS` (shared validator allow-list). |
-| `index.ts`          | Barrel re-exports from all the above files.                                                                                                                                                           |
+| File                | Purpose                                                                                                                                                                        |
+| ------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `ThemeProvider.tsx` | React component, `ThemeContext`, `InternalContext` (nesting detection), `useTheme` hook. Injects `--openui-*` vars into `<head>` via `useInsertionEffect`.                     |
+| `types.ts`          | TypeScript interfaces: `Theme`, `ColorTheme`, `LayoutTheme`, `TypographyTheme`, `EffectTheme`, `ChartColorPalette`.                                                            |
+| `defaultTheme.ts`   | Builds `defaultLightTheme` / `defaultDarkTheme` (frozen) from the swatch system. Contains `createColorTheme()` and all layout/typography/shadow defaults.                      |
+| `swatches.ts`       | 18 oklch color families x 14 shades. Exports `swatch()`, `withAlpha()`, `swatchToken()`, `swatchTokens`.                                                                       |
+| `utils.ts`          | Exports `camelToKebab`, `themeToCssVars`, `createTheme` (dev-mode typo detection with Levenshtein distance suggestions), and `KNOWN_THEME_KEYS` (shared validator allow-list). |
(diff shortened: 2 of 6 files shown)


packages/react-ui/tsconfig.test.json
ankit-thesys, line 1: what is this ?

ankit-thesys: The shared ESLint config (root `eslint.config.cjs:14-18`) type-checks `**/__tests__/**` files against `./tsconfig.test.json`, resolved per package. react-ui's main `tsconfig.json` can't serve that role — it's the build config (`rootDir: src`, `outDir: dist`) and explicitly excludes `src/**/__tests__` so tests never compile into `dist/`. This PR adds react-ui's first test, so without this file `lint:check` hard-errors: `Parsing error: TS5012: Cannot read file '.../tsconfig.test.json': ENOENT` (verified by deleting it and running eslint). Contents are just `extends ./tsconfig.json` + `noEmit` + include all of src — a copy of react-headless's existing `tsconfig.test.json`, which exists there for the same reason.

i-subham23 (approved):

PR #729, opened 2026-07-07
Nested Schema Validation & Type mismatch check

AbhinRustagi:
# Schema validation for component props: validate, repair, and prune instead of drop-or-ignore

## What this PR does

Previously the parser only checked that required top-level props were present — nested data was never validated, and a wrong-typed value flowed straight to the renderer. This PR adds **recursive JSON-Schema validation** of all component props (objects, arrays, scalar/enum leaves, and component elements in data positions), and applies one consistent **resolution rule** to every invalid value instead of either ignoring it or blanking the UI:

```
invalid value at an edge:
  1. schema has a `default`  →  substitute it (error still reported)
  2. edge is REQUIRED        →  parent becomes invalid; same rule one level up
  3. edge is OPTIONAL        →  prune the value
array items: always pruned  →  one bad row never takes down the table
```

At the outermost edge (component param), "propagate" means **drop the component**. Net effect: a deep violation can only blank the root if *every* edge on the path is required and none is an array item — i.e. only when the tree genuinely cannot satisfy the schema. LLM output degrades gracefully; the error always reaches the host's `onError` for the correction loop.

## Cases covered

| # | Case | Handling |
|---|------|----------|
| **Top-level args** | | |
| 1 | Required arg, type/enum mismatch, default provided | Substitute default, keep error |
| 2 | Required arg, type/enum mismatch, no default | **Drop component** |
| 3 | Optional arg, mismatch, default provided | Substitute default, keep error |
| 4 | Optional arg, mismatch, no default | Drop the arg (prop pruned) |
| 5 | Required arg absent/null, default provided | Fill silently, no error *(pre-existing)* |
| 6 | Required arg absent/null, no default | Error + drop component *(pre-existing)* |
| 7 | Excess positional args | Reported, extras dropped *(pre-existing)* |
| **Nested objects** | | |
| 8 | Required child key invalid, child has default | Substitute child default; parent survives |
| 9 | Required child key invalid, no default | Parent invalid → rule repeats one level up |
| 10 | Optional child key invalid, default provided | Substitute default |
| 11 | Optional child key invalid, no default | Prune the key; parent survives |
| 12 | Required child key absent/null, default provided | Fill silently, no error |
| 13 | Required child key absent/null, no default | Error (`missing-required`/`null-required`) + parent invalid |
| 14 | Wrong container kind (scalar/array where object expected, etc.) | `type-mismatch` → resolution rule |
| **Arrays** | | |
| 15 | Any invalid item (bad type, missing required key, misplaced component) | **Item pruned, siblings kept** — arrays never invalidate; error paths keep original indices |
| 16 | Non-array where array expected | `type-mismatch` → resolution rule |
| **Components in data positions** | | |
| 17 | Element where a plain-data shape is declared (e.g. `HorizontalBarChart(Card([]), …)`) | `type-mismatch` (*"expects array but got component Card"*) → resolution rule |
| 18 | Element in a component slot (`$ref`/`anyOf`) | Skipped — slot membership deferred (see below) |
| 19 | Misplaced component's own args | Still validated independently; both errors surface |
| **Streaming** | | |
| 20 | Required-key checks mid-stream | Deferred while partial; fire on completion |
| 21 | Enum membership mid-stream (e.g. `Badge("prim` cut mid-token) | Deferred — a partial literal may still complete to a valid member; judged at completion |
| 22 | Scalar type checks mid-stream | Active — a wrong scalar type can't become right by streaming more |
| **Conservative skips** (unchecked by design) | | |
| 23 | Composite schemas (`$ref`/`anyOf`/`oneOf`/`allOf`) for plain values | Skipped |
| 24 | Dynamic values (`$var`, builtins) | Skipped — resolve at runtime |
| 25 | Malformed schema fragments (unknown `type`, tuple `items`, non-array `required`) | Skipped |
| 26 | Schema `default` values themselves | Trusted, not validated |

## Error reporting

- New `type-mismatch` error code, `expects X but got Y` shape everywhere (incl. enum lists: `expects one of ["info", …] but got "inf"`).
- All error messages generated from one template switch (`ValidationIssue` → `validationMessage`) — no per-call-site drift. **Note:** top-level missing-required messages now use JSON-pointer paths (`"/theme"` instead of `"theme"`); match on `.code`/`.path`, not message text.
- **Messages are self-sufficient.** Type mismatches inline the expected type/enum list; top-level required violations inline the typed component signature (`… — signature: Callout(variant*: "info"|"warning"|…, title*: string, …)`); unknown components inline the full catalog (`… Available components: Card, TextContent, …`).
- **`enrichErrors` is deprecated** (unchanged otherwise — reverted to its `main` behavior). Since `ValidationError.message` now carries everything the hints did, consumers should read `result.meta.errors` directly — match on `.code`/`.path`, display `.message`. The pass keeps working during the transition and will be removed in a future major.

## Implementation notes

- `validation.ts` (new): shape-dispatched validators (`validateElementPosition` / `Object` / `Array` / `Leaf`) + shared `resolveInvalidValue` edge rule + issue-based error reporting.
- `materialize.ts` applies the same edge rule at the param boundary; `MaterializeCtx` moved to `types.ts` to break the import cycle.
- Pruning mutates in place — safe because `Obj`/`Arr` materialize fresh per parse; the streaming parser's cached AST is never touched.
- Also in this PR: `compileSchema` retains each param's schema fragment (`ParamDef.schema`), low-level parsing pipeline exports (`tokenize`, `parseExpression`, `split`, `autoClose`, `walkAST`, `Token`), and a prompt rule requiring filter `$bindings` to appear in Query args.

## Deferred (explicit non-goals here)

- **`$ref` resolution** — data behind `$ref` (e.g. chart `Series`) is still unchecked; biggest coverage gap, planned next.
- **Slot membership** — "is this component allowed in this children slot" (helper existed, backed out to keep scope).
- Per-statement streaming `partial` (checks currently gate on whole-document completeness); `additionalProperties`; constraint keywords (`minItems`, `pattern`, …).

## Tests

20 consolidated cases in `materialize.validation.test.ts`, one per behavior family: recursive descent & reporting, the edge rule (each row of the table above), components in data slots, conservative skips, streaming gates. Full suite: 98/98.


packages/lang-core/src/parser/validation.ts
```diff
@@ -0,0 +1,173 @@
...
+/**
+ * Check a materialized value against a scalar leaf's declared type/enum.
+ * Returns a human-readable {expected, actual} on mismatch, or null when it
+ * passes or isn't checkable. Values reaching this point are concrete literals
+ * (dynamic/element/null values are filtered out by validateSchemaValue), so a
+ * declared scalar type also flags compound values (object/array). Enum
+ * membership is only checked for scalar literals — `includes` compares
+ * non-scalar enum members by reference, which could never match.
+ */
+function describeTypeMismatch(
```
Aditya-thesys: should we move this also to enrich-error file?

Aditya-thesys (commented):
Update the react-lang also to not use enrich errors

Aditya-thesys:
Also bump the react lang version also please

Aditya-thesys (dismissed):

Aditya-thesys (dismissed):

Aditya-thesys (approved):

PR #730, opened 2026-07-07
Render markdown in Vue chat TextContent

### LINKED ISSUE #728: Vue chat TextContent claims markdown support but renders markdown literally
## Summary

The Vue chat example tells the model that `TextContent` supports markdown, but the local Vue renderer displays markdown markers literally.

Observed while running PR #629 locally, but this appears to be an example/component-library mismatch rather than part of the PR's `packages/vue-lang` query/mutation scope.

## Repro

1. Run `examples/vue-chat`.
2. Ask for a response likely to include formatted text, for example stock information.
3. The assistant response renders literal markdown such as `**NVDA**` and `**$877.33**` instead of bold text.

## Expected

Markdown inside `TextContent("...")` should render as formatted text, or the Vue chat prompt/library should not claim markdown support.

## Actual

Markdown is shown literally.

## Relevant code

- `examples/vue-chat/lib/library.ts` describes `TextContent` as: "Displays a block of text. Supports markdown formatting within the string."
- `examples/vue-chat/lib/library.ts` also instructs: "Use TextContent for all text output. You can use markdown within the text string."
- `examples/vue-chat/components/openui/TextContent.vue` renders the text using plain Vue interpolation: `{{ props.text ?? "" }}`, which escapes and displays markdown syntax literally.

## Likely fix options

- Add a markdown renderer to the Vue chat `TextContent.vue` implementation, with sanitization appropriate for rendered HTML.
- Or update the Vue chat prompt/library description so it does not instruct the model to emit markdown in `TextContent`.

## Notes

`@openuidev/vue-lang` appears to be the runtime/bindings layer (`Renderer`, `defineComponent`, `createLibrary`, composables). The Vue chat component library is local to the example. The richer prebuilt markdown component exists on the React side as `@openuidev/react-ui`'s `MarkDownRenderer`, but there is no equivalent Vue UI package in this example.

shogun444:
Fixes #728

## What

The Vue chat example claims that `TextContent` supports markdown, but the component rendered text using plain Vue interpolation, causing markdown syntax (e.g. `**NVDA**`) to appear literally instead of being formatted.

This PR updates `TextContent` to render markdown as the prompt already describes.

## Changes

- Parse `TextContent` using `marked`
- Sanitize the generated HTML with `DOMPurify`
- Render the sanitized HTML using `v-html`
- Add `marked` and `dompurify` as direct dependencies for the Vue chat example

## Why

The issue proposed two possible fixes:

- Render markdown correctly
- Remove markdown support from the prompt

This PR chooses the first approach so the implementation matches the documented behavior instead of reducing functionality.

The change is intentionally scoped to `examples/vue-chat` only. No changes were made to `@openuidev/vue-lang` since the runtime is functioning correctly; the issue was limited to the example's rendering implementation.

## Test Plan

- [x] Plain text rendering is unchanged
- [x] HTML is sanitized before rendering (`DOMPurify`)

## Checklist

- [x] I linked a related issue
- [x] No documentation changes required

### DIFF
--- a/examples/vue-chat/components/openui/TextContent.vue
+++ b/examples/vue-chat/components/openui/TextContent.vue
@@ -1,7 +1,126 @@
 <script setup lang="ts">
-defineProps<{ props: { text?: string } }>();
+import DOMPurify from "dompurify";
+import { marked, Renderer } from "marked";
+import { computed } from "vue";
+
+const { props } = defineProps<{ props: { text?: string } }>();
+
+const renderer = new Renderer();
+renderer.html = ({ text }: { text: string }) => {
+  return text.replace(/&/g, "&amp;").replace(/</g, "&lt;").replace(/>/g, "&gt;");
+};
+
+const html = computed(() => {
+  const raw = marked.parse(props.text ?? "", { renderer, async: false });
+  return DOMPurify.sanitize(raw, {
+    ALLOWED_TAGS: [
+      "p", "br", "strong", "em", "code", "pre", "a", "ul", "ol", "li",
+      "h1", "h2", "h3", "h4", "h5", "h6", "blockquote", "hr",
+      "table", "thead", "tbody", "tr", "th", "td",
+    ],
+    ALLOWED_ATTR: ["href", "target", "rel"],
+  });
+});
 </script>
 
 <template>
-  <p class="text-sm leading-relaxed text-zinc-700 dark:text-zinc-300">{{ props.text ?? "" }}</p>
+  <div class="markdown-content text-sm leading-relaxed text-zinc-700 dark:text-zinc-300" v-html="html"></div>
 </template>
+
+<style scoped>
+.markdown-content :deep(h1) {
+  font-size: 1.5em;
+  font-weight: 700;
+  margin: 0.67em 0;
+  line-height: 1.2;
+}
+.markdown-content :deep(h2) {
+  font-size: 1.25em;
+  font-weight: 600;
+  margin: 0.75em 0;
+  line-height: 1.3;
+}
+.markdown-content :deep(h3) {
+  font-size: 1.1em;
+  font-weight: 600;
+  margin: 0.83em 0;
+  line-height: 1.4;
+}
+.markdown-content :deep(h4),
+.markdown-content :deep(h5),
+.markdown-content :deep(h6) {
+  font-size: 1em;
+  font-weight: 600;
+  margin: 1em 0;
+  line-height: 1.4;
+}
+.markdown-content :deep(p) {
+  margin: 0.5em 0;
+}
+.markdown-content :deep(strong) {
+  font-weight: 600;
+}
+.markdown-content :deep(em) {
+  font-style: italic;
+}
+.markdown-content :deep(code) {
+  font-size: 0.875em;
+  padding: 0.15em 0.4em;
+  border-radius: 0.25rem;
+  background: rgb(228 228 231 / 1);
+}
+.markdown-content :deep(pre) {
+  overflow-x: auto;
+  padding: 0.75rem;
+  border-radius: 0.5rem;
+  background: rgb(24 24 27 / 1);
+  color: rgb(212 212 216 / 1);
+  margin: 0.75em 0;
+}
+.markdown-content :deep(pre code) {
+  padding: 0;
+  background: transparent;
+  border-radius: 0;
+}
+.markdown-content :deep(ul),
+.markdown-content :deep(ol) {
+  padding-left: 1.5em;
+  margin: 0.5em 0;
+}
+.markdown-content :deep(ul) {
+  list-style-type: disc;
+}
+.markdown-content :deep(ol) {
+  list-style-type: decimal;
+}
+.markdown-content :deep(li) {
+  margin: 0.25em 0;
+}
+.markdown-content :deep(a) {
+  color: rgb(99 102 241 / 1);
+  text-decoration: underline;
+}
+.markdown-content :deep(blockquote) {
+  border-left: 3px solid rgb(212 212 216 / 0.4);
+  padding-left: 0.75em;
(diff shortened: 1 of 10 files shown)


AlexQuidditch:
I would be happy to see this feature merged asap 🙏

examples/vue-chat/components/openui/TextContent.vue
```diff
@@ -1,7 +1,16 @@
...
-defineProps<{ props: { text?: string } }>();
+import DOMPurify from "dompurify";
+import { marked } from "marked";
+import { computed } from "vue";
+
+const { props } = defineProps<{ props: { text?: string } }>();
+
+const html = computed(() => {
+  const raw = marked.parse(props.text ?? "", { async: false });
+  return DOMPurify.sanitize(raw);
```
vishxrad: [P2] Please disable or escape raw HTML before sanitizing. marked.parse() preserves embedded HTML, while the default DOMPurify policy still permits forms, inputs, external form actions and images, and fixed-position inline styles; a model response can therefore render UI outside the component library controls. HTML-looking plain text such as <script> also loses content instead of being displayed. Prefer treating Marked HTML tokens as text, then keep DOMPurify with a Markdown-specific tag and attribute allowlist.

examples/vue-chat/components/openui/TextContent.vue
```diff
@@ -1,7 +1,16 @@
...
+});
 </script>
 
 <template>
-  <p class="text-sm leading-relaxed text-zinc-700 dark:text-zinc-300">{{ props.text ?? "" }}</p>
+  <p class="text-sm leading-relaxed text-zinc-700 dark:text-zinc-300" v-html="html"></p>
```
vishxrad: [P2] marked.parse() emits block elements; even plain text becomes <p>...</p>, so injecting the result into an outer <p> creates invalid nested or block markup. Tailwind preflight also removes list markers, heading and link styling, and default spacing, leaving most block Markdown visually unformatted. If this component only needs inline formatting, use marked.parseInline(); if it should support full Markdown, use a <div> host and add descendant Markdown styles.

vishxrad (changes requested):
Thanks for fixing #728. I am requesting changes for two issues before merge: model-produced text should stay within a Markdown-only rendering surface, and the component should consistently support either inline Markdown or correctly structured and styled block Markdown. The dependency and lockfile changes look good, and both CI jobs pass.

AlexQuidditch:
@shogun444 maybe https://github.com/Simon-He95/markstream-vue would be more suitable for this case

shogun444:
@AlexQuidditch  Thanks for the suggestion! markstream-vue looks great for streaming use cases. For now I've addressed the @vishxrad  feedback with a minimal fix (HTML escaping, proper block structure, scoped markdown styles). If there's interest in a fuller streaming markdown solution for the Vue chat example, that could be a follow-up.

Sorry for the late PR.

AlexQuidditch:
@shogun444 

I've implemented `markstream-vue` into my fork with `vue-headless` package. It's working great.

Screenshot 2026-07-15 at 13 12 46

vishxrad (changes requested):
Requesting changes on head `49818825`.

- **[P1] `TextContent.vue:11` — all `TextContent` output currently throws.** With locked `marked@17.0.5`, the per-call `renderer` option replaces the complete renderer; this partial `{ html() {} }` object is not merged with the defaults. Even `marked.parse("hello", options)` calls the missing `paragraph()` method and throws `TypeError: this.renderer.paragraph is not a function`. In the running Vue example, submitting `What is Vue?` renders headings and buttons but no body copy, and the DOM contains zero `.markdown-content` nodes. Use a complete `Renderer` instance with `html` overridden, or configure a dedicated `Marked` instance through `.use({ renderer: ... })`.

- **[P2] `TextContent.vue:29` — dark-mode body text regresses.** The new wrapper removes `text-sm leading-relaxed text-zinc-700 dark:text-zinc-300`. Neither `Card` nor `Stack` supplies a body-text color, while the page uses `dark:bg-zinc-950`; after fixing the parser, ordinary Markdown will inherit black and be unreadable in dark mode. Restore the base typography and light/dark text colors.

- **[P2] `TextContent.vue:20` — Markdown headings have no presentation.** `h1`–`h6` are allowed through sanitization, but there are no heading rules. Tailwind preflight resets heading size and weight to `inherit`, so headings render like body text. Add scoped heading typography and spacing.

Please add regression coverage for plain text, bold/inline-code Markdown, and escaped raw HTML. `markstream-vue` looks worth evaluating separately for streaming Markdown UX, but adopting it needs streaming-state wiring, an explicit escaped-HTML policy, CSS ordering, and bundle/visual QA; it is not required for this focused fix.

vishxrad:
@shogun444 Thanks — the runtime, security, and styling issues from my review are fixed on `e305766`.

One small follow-up remains before I can approve: the new `TextContent.test.ts` suite is currently not runnable. The latest commit removed the `test` script and the required `vitest`, `@vitejs/plugin-vue`, `@vue/test-utils`, and `jsdom` dev dependencies. As a result, `pnpm --filter vue-chat exec vitest run` fails with `Command "vitest" not found`, and the current CI jobs only build the examples.

Could you please:

1. Restore the test dev dependencies and update `pnpm-lock.yaml`.
2. Add a script such as `"test": "nuxt prepare && vitest run"` (`nuxt prepare` is needed for `.nuxt/tsconfig.json` on a clean checkout).
3. Run `pnpm --filter vue-chat test` in CI.

I verified that all four tests pass once the dependencies are installed and Nuxt is prepared. After that, I’m good to approve.

shogun444:
Adding multiple providers in the vue-chat could be a follow up similar to the openui-chat so user's can bring in their own API key and play with it.
Rightnow, it only supports OpenAI models.

@vishxrad  do you want any other changes?

PR #737, opened 2026-07-08
feat(react-ui): SSR-correct theming — rendered style tag, data-openui-mode contract, system mode, scheme-pinned defaults

### LINKED ISSUE #736: Dark only apps render the stock light theme on light scheme devices because there is no scheme pinned CSS export
## Summary

Every dark token that `@openuidev/react-ui` ships is gated behind `@media (prefers-color-scheme: dark)`. This is true for `./index.css`, `./layered/styles/index.css`, and `./defaults.css`. There is no shipped stylesheet that pins a single scheme statically.

An app that forces dark mode with `<ThemeProvider mode="dark">` therefore renders the stock light theme on any device whose system scheme is light, for every token the app has not personally overridden. This happens in every situation where the runtime style injection from ThemeProvider is not in effect, which includes the static first paint, server rendered HTML before hydration, hydration failures, and pages with JavaScript disabled. The mode prop only affects the runtime context and the injected style tag. It never changes what the static CSS resolves to.

## Real world impact

A downstream dark only consumer hit this in production. The app forced dark mode and bridged about 25 brand tokens with its own unlayered `:root` overrides, which is the documented theming contract. On a desktop with a dark system scheme everything looked correct. On a phone with a light system scheme the chat interface rendered as a full stock light UI inside an otherwise dark app, because the chat surfaces read tokens the bridge did not override, and those tokens resolved from the light `:root` block since the dark values sit behind the media query.

The workaround that consumer landed on was to scrape the full dark token block out of the dist CSS with a script at every upgrade, paste roughly 21 KB of duplicated tokens into their own repo, and guard it with a drift test. That is exactly the kind of thing a first class export should replace.

## Proposal

Emit two new scheme pinned stylesheets from the same generator that already produces `openui-defaults.css` (`src/scripts/generate-css-utils.ts`), and add them to the exports map:

1. `./defaults-dark.css` containing a plain unlayered `:root` block with the full dark token set and no media query.
2. `./defaults-light.css` containing the same for light, since light only apps have the mirrored problem on dark scheme devices.

Usage for a dark only app then becomes one extra import and zero scraping:

```css
@layer theme, base, openui, components, utilities;
@import "@openuidev/react-ui/layered/styles/index.css";
@import "@openuidev/react-ui/defaults-dark.css";
/* the consumer's own :root brand token overrides follow and win by source order */
```

The generator already has `defaultDarkTheme` in scope and already iterates it for the media gated block, so the new artifacts come from the token source of truth rather than from transforming the output CSS. The build pipeline picks new `.scss` files in `src` up automatically. Alongside the generator change this needs a copy step in `cp-css.js`, two entries in the exports map, and a short docs section explaining the single scheme setup.

A fuller investigation with exact source references lives in `docs/dark-scheme-exports-investigation.md`.

ankit-thesys:
Fixes #736

## Problem

The theme reached the page only through a client-side `useInsertionEffect`, so server-rendered HTML painted from the static CSS alone — and that CSS keeps every dark token behind `@media (prefers-color-scheme: dark)`. Measured consequences (Playwright, pre-hydration vs post-hydration):

* An app forcing a mode (`<ThemeProvider mode="dark">`) painted the device's scheme first, then flipped at hydration — a full-page theme flicker on every mismatched device, and permanently wrong styling wherever hydration never runs (JS disabled, hydration failure).
* Apps passing custom `lightTheme`/`darkTheme` tokens flashed stock values → brand values at hydration.
* A stored user preference ("user chose dark, device is light") was unfixable: no stylesheet could express it and the runtime injection arrived only after hydration.

## Change

**ThemeProvider renders its rules as a `<style>` element in its output** (replacing the effect-time injection), so SSR/streamed HTML carries the resolved theme before first paint. Explicit `mode="light" | "dark"` pins one resolved token set, exactly as the injected tag did post-hydration.

**New `mode="system"`** follows the device scheme and resolves the context `mode` client-side (attribute first, then `matchMedia`, via `useSyncExternalStore`). The provider's tag carries only what the static CSS cannot know: with no custom themes a top-level system provider renders **no style tag at all** (the defaults stylesheet already implements system-following + the attribute override); custom tokens emit in the same light/media-dark/attribute structure carrying only overridden values; scoped/nested providers keep full sets so a subtree can diverge from an ancestor's pinned tokens. `useTheme().mode` always surfaces the resolved `"light"`/`"dark"`. The default mode is unchanged (`"light"`) — existing apps keep their exact behavior; whether `"system"` should become the default is left as a follow-up decision.

**`data-openui-mode` attribute contract in the static CSS.** The generated defaults now also emit `[data-openui-mode="light"]` / `[data-openui-mode="dark"]` full-token blocks after the media-gated block, so a pinned scheme is expressible in plain HTML (`<html data-openui-mode="dark">` or any wrapper element) with zero JavaScript. `check-css-artifacts.js` asserts the blocks' presence, order, and token parity in both trees.

**New `ThemeScript` export** — a tiny inline script for `<head>` that applies a stored localStorage preference (`"openui-theme"` key by default) to `<html>` before first paint; pairs with `mode="system"`. Storage key is serialized inert (JSON + `<` escaping). The cookie-based server-side variant is documented in the README.

**Scheme-pinned stylesheets** (`./defaults-dark.css`, `./defaults-light.css`) ship as before for CSS-only single-scheme setups; the attribute contract is the general mechanism.

**Example**: openui-cloud drops its hand-rolled system-theme hook (which returned `"light"` during SSR) for `theme={{ mode: "system" }}`.

## Not in scope

Components that bake the context mode into their own markup (`ChartContainer`'s inline token re-pins, `CodeBlock`/`MarkDownRenderer`'s mode-picked highlight style objects) can still briefly restyle at hydration on a mode-mismatched device. That component-level flicker is accepted for now — the page itself stays scheme-stable — and can be addressed separately if it ever matters.

## Verification

`pnpm build` (including the extended `check:css`), `tsc --noEmit`, eslint, and prettier all pass; `pnpm test` passes with 10 new `renderToString` SSR tests (rendered-tag shapes for pinned/system/nested modes, custom-token placement, ThemeScript escaping incl. a hostile storage key).

Playwright pre-hydration (JS disabled) vs post-hydration probes on the example, all zero-delta after this change:

| Scenario | Before | After |
|---|---|---|
| `mode="system"`, either device scheme | no flicker | no flicker |
| Forced dark, light-scheme device | full-page light→dark flip at hydration | dark from the first SSR frame, JS disabled |
| Stored dark pref + `ThemeScript`, light device | not expressible | dark before hydration, verified with hydration bundles blocked |
| Server HTML | no theme present | provider style tag in SSR output |


packages/react-ui/src/components/ThemeProvider/ThemeProvider.tsx
```diff
@@ -232,28 +287,45 @@ export const ThemeProvider = ({
...
+  }, [mode, resolvedLightTheme, resolvedDarkTheme, styleSelector, portalClassName]);
 
   // ---------------------------------------------------------------------------
   // Render
   // ---------------------------------------------------------------------------
   return (
     <InternalContext.Provider value={internalValue}>
       <ThemeContext.Provider value={contextValue}>
+        {/* Rendered before children so streamed CSS parses before content paints. */}
+        <style data-openui-theme={id} dangerouslySetInnerHTML={{ __html: styleContent }} />
```
ankit-thesys: do we need this.

ankit-thesys: i mean this is a hackly looks code here

packages/react-ui/src/components/ThemeProvider/ThemeScript.tsx
```diff
@@ -0,0 +1,38 @@
...
+ * </html>
+ * ```
+ */
+export const ThemeScript = ({ storageKey = "openui-theme" }: ThemeScriptProps) => {
+  // JSON.stringify + <-escaping keeps any storageKey inert inside the
+  // inline script: it stays a JS string literal and cannot close the tag early.
+  const key = JSON.stringify(storageKey).replace(/</g, "\\u003c");
+  const code = `(function(){try{var v=localStorage.getItem(${key});if(v==="light"||v==="dark")document.documentElement.setAttribute("data-openui-mode",v)}catch(e){}})()`;
+  return <script dangerouslySetInnerHTML={{ __html: code }} />;
+};
```
ankit-thesys, line 52: i don't undersand this
write human redable code

packages/react-ui/src/scripts/generate-css-utils.ts
```diff
@@ -122,24 +122,37 @@ fs.writeFileSync(outputPath, lines.join("\n"));
...
+      target.push(`${indent}  ${sectionHeaders[key]}`);
+      target.push("");
+    }
+    const kebab = camelToKebab(key);
+    target.push(`${indent}  --openui-${kebab}: ${value};`);
+  }
+
+  target.push(`${indent}}`);
+}
+
```
ankit-thesys, line 149: explain this block

packages/react-ui/src/components/ThemeProvider/ThemeProvider.tsx
ankit-thesys: Yes, both halves are needed, though this thread led to a real improvement — details:

**Why a rendered `<style>` element at all:** this is the fix itself. The old `useInsertionEffect` injection only runs on the client, so server-rendered HTML had no theme and painted from the media-query CSS alone (that's the flicker). Rendering the tag as part of the output makes it stream with the SSR HTML, so the correct tokens exist before first paint.

**Why `dangerouslySetInnerHTML`:** React HTML-escapes normal text children — `[data-openui-mode="dark"]` would ship as `[data-openui-mode=&quot;dark&quot;]`, which is broken CSS. Raw text inside `<style>` requires `dangerouslySetInnerHTML`; it's the standard pattern for SSR'd style tags and `ChartStyle` in `Charts.tsx` has used it here for a long time. The content is our own generated token strings, never user HTML.

**The improvement (e48b3e47):** the tag now only renders when it has something the static CSS can't provide — a top-level `mode="system"` provider with no custom themes renders no tag at all, and custom themes emit only their overridden tokens. That removed ~53KB of duplicated inline CSS per page and most of what made this block look heavy.

packages/react-ui/src/components/ThemeProvider/ThemeScript.tsx
ankit-thesys: Agreed — rewritten in e48b3e47. The logic is now a plain named function (`applyStoredMode`: read localStorage, and if the value is `"light"` or `"dark"`, set `data-openui-mode` on `<html>`) and the inline script is just `(${String(applyStoredMode)})(key)` — so the source reads like normal code instead of a hand-minified string.

The one remaining bit of ceremony is the key serialization: `JSON.stringify` keeps the storage key a JS string literal, and escaping `<` as `\u003c` stops a key containing `</script>` from closing the tag early (there's a test that feeds it a hostile key to lock that in).

packages/react-ui/src/scripts/generate-css-utils.ts
ankit-thesys: This is the shared emitter every generated token block goes through. Given a selector, the theme entries, and an indent, it prints:

```css
selector {
  // Surface / Background Colors

  --openui-background: …;
  --openui-foreground: …;
  …
}
```

walking the entries in order and inserting the section-header comments (`// Text Colors`, `// Spacing`, …) with a blank line above each. The one subtle line is the `target.length > blockStart + 1` guard: it skips the blank line for the *first* header only, so a block never opens with a stray empty line right after the brace.

Four call sites reuse it — the `:root` light block, the media-gated dark block, the two `[data-openui-mode]` blocks, and the scheme-pinned defaults files — which is why it takes selector/indent as parameters; previously that loop existed as four near-identical copies.

PR #742, opened 2026-07-08
Front-load OpenUI Cloud setup in CLI

zahlekhan:
Summary:
- Front-load create-flow choices before dependency install
- Add ranked create funnel telemetry, including scaffold/dependency outcomes
- Update CLI docs for AI setup and existing create flags

Checks:
- tsc -p packages/openui-cli
- eslint ./src --ignore-pattern 'src/templates/**'
- prettier --check './src/**' '!./src/templates/**'
- quick smoke creates with --no-install/--no-skill


packages/openui-cli/src/commands/create-app.ts
```diff
@@ -22,6 +22,54 @@ export interface CreateAppOptions {
...
+  authMethod?: CloudAuthMethod | "apikey-flag";
+  authSucceeded?: boolean;
+};
+
+const createFunnel = {
+  funnel: "cli_create",
+  funnel_version: "frontloaded_cloud_setup_v1",
+} as const;
+
+const createFunnelSteps = {
```
AbhinRustagi: Can you move all telemetry stuff to its file or a `telemetry-utils.ts` file?

AbhinRustagi (changes requested):

packages/openui-cli/src/commands/create-app.ts
AbhinRustagi: Also so core handlers are not noisy

AbhinRustagi (approved):

PR #744, opened 2026-07-08
fix(react-lang): stop ElementErrorBoundary from crashing the host with a commit-phase insertBefore error

### LINKED ISSUE #727: Renderer: commit-phase `insertBefore` NotFoundError from ElementErrorBoundary re-presenting stale children (+ index-keyed children)
## Summary

`@openuidev/react-lang@0.2.1`'s `Renderer` can crash the host app with a React commit-phase error while re-rendering or streaming content:

```
NotFoundError: Failed to execute 'insertBefore' on 'Node':
The node before which the new node is to be inserted is not a child of this node.
```

The error is thrown deep in React's commit (mutation) phase, so it **cannot be caught by a nested error boundary** in the host app — it escalates to the app root and unmounts the whole subtree. We hit it in production in a streaming chat UI, and also when a sibling re-render forces an already-rendered `Renderer` subtree to re-render.

## Root cause

Two interacting behaviors in the renderer (line refs from the published `dist/index.mjs`, v0.2.1):

**1. `ElementErrorBoundary` re-presents stale children on error.** On `hasError`, `render()` returns `this.lastValidChildren` — React element objects captured from a *previous* render (`componentDidMount`/`componentDidUpdate` stash `this.props.children`). Those elements correspond to DOM nodes React has already reconciled/moved. Re-inserting them makes React's fiber tree disagree with the live DOM:

```js
// dist/index.mjs ~L443-471
componentDidUpdate(prevProps) {
  if (!this.state.hasError) this.lastValidChildren = this.props.children;
  if (this.state.hasError && prevProps.children !== this.props.children) this.setState({ hasError: false });
}
render() {
  if (this.state.hasError) return this.lastValidChildren; // <-- stale element tree
  return this.props.children;
}
```

**2. Array children are keyed by array index.** `renderDeep` renders arrays as index-keyed fragments:

```js
// dist/index.mjs ~L482
if (Array.isArray(value)) return value.map((v, i) => jsx(Fragment, { children: renderDeep(v) }, i));
```

When the parsed structure at a given index changes type between renders (which happens constantly while streaming a partial DSL, or when a child throws and the boundary swaps in stale children and back), index keys make React reconcile mismatched subtrees. Combined with (1), React's DOM bookkeeping desyncs and `insertBefore(newNode, referenceNode)` fails because `referenceNode` is no longer a child of the expected parent.

## Reproduction (essence)

1. Render a moderately deep tree (e.g. a Card with an array of child elements) via `<Renderer>`.
2. Cause a child component to throw intermittently during re-render — e.g. stream a DSL so it is repeatedly re-parsed while partially formed, so a component occasionally receives malformed/partial props and throws.
3. `ElementErrorBoundary` catches, returns stale `lastValidChildren`, the next valid render swaps back — at index-keyed positions whose types shift.
4. Intermittently: `NotFoundError: Failed to execute 'insertBefore' ...` in React's commit phase, which escapes to the app root.

## Suggested fixes

- **`ElementErrorBoundary` on error should NOT return stale element instances.** Return `null` (or a caller-supplied fallback), or reset via a `key` bump so the subtree cleanly remounts. Re-inserting elements whose DOM React already moved is the core hazard.
- **Prefer stable keys over array index** in `renderDeep` where a stable identity is derivable from the node, so type changes at a position remount cleanly rather than reconciling mismatched subtrees.

## Workaround we're using

A local `pnpm patch` changing `ElementErrorBoundary.render()` to `return null` on error. This eliminated the crash across streaming and re-render triggers, at the cost of losing the "show last good state" behavior for a single element for one frame (acceptable vs. an app crash).

## Environment

- `@openuidev/react-lang` 0.2.1
- React 19.2, Next.js 16

serhiizghama:
Closes #727.

`ElementErrorBoundary` catches a render error and then returns `this.lastValidChildren` — element instances captured from an earlier render. React has already reconciled and moved the DOM for those elements, so re-inserting them desyncs the fiber tree from the live DOM and throws `NotFoundError: Failed to execute 'insertBefore' on 'Node'` in the commit phase. Because it's a mutation-phase error it can't be caught by a nested boundary — it escalates to the app root and unmounts the whole subtree. It shows up in streaming chat UIs and whenever a sibling re-render forces an already-rendered `Renderer` subtree to re-render.

The fix is to render `null` on error rather than re-presenting the stale children. The boundary still auto-recovers as soon as new valid children arrive (the existing `componentDidUpdate` reset), so a transient streaming error clears on the next update instead of crashing. This matches the workaround the reporter verified. The trade-off is losing the "show last good state" for the one frame an element is broken, which seems well worth not tearing down the app — but happy to reconsider if you'd rather keep last-good-state via a keyed remount instead.

The package had no test setup, so I added a jsdom vitest config and a regression test driving the boundary through error → recovery: it asserts the boundary renders nothing after a child throws (not the stale subtree) and recovers when valid children return. Confirmed it fails against the old `return this.lastValidChildren`. typecheck/eslint/prettier clean.

### DIFF
--- a/packages/react-lang/package.json
+++ b/packages/react-lang/package.json
@@ -81,6 +81,8 @@
   "devDependencies": {
     "@modelcontextprotocol/sdk": "^1.27.1",
     "@types/react": "catalog:",
+    "@types/react-dom": "catalog:",
+    "react-dom": "catalog:",
     "vitest": "^4.0.18"
   }
 }
--- a/packages/react-lang/src/Renderer.tsx
+++ b/packages/react-lang/src/Renderer.tsx
@@ -67,14 +67,17 @@ interface ErrorBoundaryState {
 }
 
 /**
- * Error boundary that intentionally shows the last successfully rendered
- * children when a render error occurs. This "show last good state" behavior
- * prevents the UI from going blank during streaming or transient evaluation
- * errors, and auto-recovers when new valid children arrive.
+ * Error boundary that isolates a single rendered element: when a child throws
+ * during render it renders nothing for that element (instead of tearing down
+ * the whole tree) and auto-recovers as soon as new valid children arrive, e.g.
+ * the next streaming update.
+ *
+ * It deliberately does NOT re-present the previously rendered children on error.
+ * Those are element instances from an earlier render whose DOM React has already
+ * reconciled/moved, so re-inserting them desyncs the fiber tree from the live
+ * DOM and throws an uncatchable `insertBefore` error in the commit phase (#727).
  */
-class ElementErrorBoundary extends Component<ErrorBoundaryProps, ErrorBoundaryState> {
-  private lastValidChildren: React.ReactNode = null;
-
+export class ElementErrorBoundary extends Component<ErrorBoundaryProps, ErrorBoundaryState> {
   constructor(props: ErrorBoundaryProps) {
     super(props);
     this.state = { hasError: false };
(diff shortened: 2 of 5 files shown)


serhiizghama:
Closing — out of date after the Renderer changes in #893/#1004.

PR #753, opened 2026-07-12
fix(react-ui): wrap long unbreakable labels in RadioItem/CheckBoxItem

Shinyaigeek:
## Problem

When `label` (or `description`) on `RadioItem` / `CheckBoxItem` contains a long token with no break opportunities — a URL, a hyphenated identifier, a long product code — the text overflows the item container and is silently clipped by its `overflow: hidden`, so part of the text disappears on narrow (mobile-width) viewports.

Normal prose (including long CJK sentences) already wraps fine; the clipping only bites on unbreakable tokens.

**Repro:** render at ~320px width

```tsx
<CheckBoxItem label="https://example.com/very/long/unbroken/path/segment-that-cannot-wrap-anywhere-because-it-has-no-spaces-at-all-1234567890" />
```

## Screenshots

Four cases per shot: long JA prose / long EN prose / unbroken URL / short label + long description. In *before*, the third item silently loses `unbroken/path/segment-` to clipping.

### CheckBoxItem

| Before | After |
| --- | --- |
| CheckBoxItem before: URL label clipped mid-string | CheckBoxItem after: URL label fully wrapped |

### RadioItem

| Before | After |
| --- | --- |
| RadioItem before: URL label clipped mid-string | RadioItem after: URL label fully wrapped |

## Fix

- Add `overflow-wrap: anywhere` to the label and description of both components so unbreakable tokens wrap instead of being clipped. (`anywhere` — rather than `break-word` — also affects min-content sizing, so it wraps correctly inside the flex row.)
- Add `min-width: 0` to the content wrapper so it can shrink within the flex container.
- Define `.openui-radio-item-content`, which `RadioItem.tsx` already renders but had no styles for, matching the existing `.openui-checkbox-item-content`.

CSS-only; no API changes.

### DIFF
--- a/packages/react-ui/src/components/CheckBoxItem/checkBoxItem.scss
+++ b/packages/react-ui/src/components/CheckBoxItem/checkBoxItem.scss
@@ -54,6 +54,7 @@
 .openui-checkbox-item-content {
   display: flex;
   flex-direction: column;
+  min-width: 0;
 }
 
 .openui-checkbox-item-label {
@@ -61,6 +62,7 @@
   @include cssUtils.typography(primary, default);
   color: cssUtils.$text-neutral-primary;
   cursor: pointer;
+  overflow-wrap: anywhere;
   &:disabled {
     color: cssUtils.$text-neutral-tertiary;
     cursor: not-allowed;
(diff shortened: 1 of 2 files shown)


vishxrad (approved):
lgtm.

PR #800, opened 2026-07-21
Add Hook for Autofocus on Agent Interface

AbhinRustagi:
Adds a hook that auto-focuses the text input on Welcome Screen & on thread switch. Disabled for mobiles to prevent automatic keyboard pop.

Should be merged with #802, #790 and changes for background requests.

### Why not HTML `autofocus` attribute?

Composer is mounted persistently and only becomes visible via CSS, so `autofocus` won't re-fire.

Closes [TH-2390](https://linear.app/thesys/issue/TH-2390/when-starting-a-new-chat-text-input-is-not-in-focus).


ankit-thesys (approved):
Looks Good

PR #805, opened 2026-07-21
Remove stale @rolldown/binding-darwin-arm64 pin and declare Node engines

pd-builds:
## Summary

The root `package.json` pinned `@rolldown/binding-darwin-arm64` at `1.0.0-rc.16` (added in #465 as a local workaround). Rolldown installs its own platform binding via optionalDependencies, so the pin is unnecessary — and actively harmful: when pnpm skips the optional binding (e.g. on Node < 20.19, which fails the binding's engines check), rolldown falls back to the stale hoisted rc.16 binding and crashes at vite/vitest startup with an opaque `TypeError: Cannot convert undefined or null to object`.

- Remove the stale pin (lockfile entries for rc.16 removed surgically — regenerating the lockfile re-resolves unrelated peer graphs, so this diff stays at 11 lines)
- Declare `engines: { node: "^20.19.0 || >=22.12.0" }" matching vite 8 / rolldown requirements, so unsupported Node versions get a clear warning instead of the cryptic crash

## Test plan

- `pnpm install --frozen-lockfile` passes (same check CI runs)
- `pnpm test` in `packages/react-headless` (109 passed) and `packages/react-ui` (6 passed)


ankit-thesys (approved):
Reviewed: stale rc.16 binding pin removal verified against lockfile (no remaining references), engines range matches rolldown requirements, frozen-lockfile install passes after conflict resolution.

PR #809, opened 2026-07-22
fix(react-ui): collapsed-sidebar open button could become unreachable (reveal keyed to animation state)

i-subham23:
Fixes the "open sidebar icon is randomly not visible" bug and keeps the collapsed rail's logo/button hover swap.

## Problem

The collapsed rail hides the open button via the header's `--collapsed` class (binary, driven by `isCollapsedLayout`), but the hover rule that reveals it keyed off `[data-sidebar-visual-state="collapsed"]` — a four-state animation attribute (`expanded → collapsing → collapsed → expanding`) advanced by chained timeouts. During the collapse animation the attribute still reads `"collapsing"`, so hovering revealed nothing (~170ms measured on every collapse, with the button visibly fading out under the cursor); and if the timeout chain is interrupted (rapid toggling, layout switches), the attribute can strand there — a collapsed rail whose open button can never be revealed. That's the randomness.

## Change

All in `sidebar.scss`; behavior:

* **The resting face is contextual** (`:has`): with a logo configured, the logo is the resting face and the button appears on hover; with no logo the slot is empty, so the button stays visible at rest instead — the open control is always discoverable.
* **The reveal rule keys off `.openui-agent-sidebar-container--collapsed`** — the same binary `isCollapsedLayout` state as the hiding rule, set in the same render — so hidden-but-unrevealable is impossible by construction.
* **Keyboard focus reveals the swap** via `:has(:focus-visible)`, so tabbing never lands on an invisible button. (`:focus-visible` rather than `:focus-within`, so the residual focus a mouse click leaves on the toggle doesn't pin the button over the logo at rest.)
* **`@media (hover: none)` shows the open button at rest on touch devices**, which have no hover to trigger the swap.

## Test Plan

- [x] Playwright probe against the openui-cloud example, with and without a `logoUrl`:
  - rest shows the logo (or the button when no logo is set); hover swaps, unhover swaps back
  - hovering **mid-collapse-animation** (the old bug window, attribute = `"collapsing"`) keeps the button fully visible — zero dead-window samples
  - after a mouse-click collapse the rail rests on the logo (click focus doesn't pin the button)
  - Tab reveals the button and Enter expands the sidebar
  - emulated `hover: none` shows the button at rest
- [x] SCSS compiles clean; CSS artifact check passes

### DIFF
--- a/packages/react-ui/src/components/AgentInterface/sidebar.scss
+++ b/packages/react-ui/src/components/AgentInterface/sidebar.scss
@@ -207,13 +207,6 @@ $sidebar-shadow: inset -10px 0 18px -18px cssUtils.$sunk-deep;
   }
 }
 
-// On hover (collapsed), swap the logo out for the toggle button. Gated on the
-// settled `data-sidebar-visual-state="collapsed"` state so the swap only runs
-// once the collapse animation has finished — keying it on the `--collapsed`
-// class instead would let the button flash in mid-collapse, while the still-wide
-// sidebar is momentarily under the cursor. (No-op when there is no logo: the
-// button is already shown at rest via the :has() rule above — which is what
-// fixes the reported TH-2392 no-logo case.)
 .openui-agent-sidebar-container[data-sidebar-visual-state="collapsed"]:hover {
   .openui-agent-sidebar-header__logo {
     opacity: 0;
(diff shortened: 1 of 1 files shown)


ankit-thesys (approved):
LG

PR #827, opened 2026-07-23
Use GPT-5.4 and disable local compare tools

zahlekhan:
## Summary

- remove the mock web search, weather, and stock tools from the Markdown and OpenUI OSS comparison path
- use a plain streaming chat completion when serving those local comparison modes
- change the OpenUI Cloud default model from Claude Sonnet 4.6 to GPT-5.4

## Why

The compare experience describes built-in tools as a Cloud-only capability, but the Markdown and OpenUI OSS endpoint was still exposing three mock tools. Removing those tools keeps the comparison aligned with the product behavior it presents. Using GPT-5.4 as the Cloud default also lets the surfaces be evaluated with the requested model.

## Impact

Markdown and OpenUI OSS no longer invoke local function tools. OpenUI Cloud keeps its existing built-in tools and now starts with `openai/gpt-5.4`.

## Validation

- `pnpm exec eslint app/api/chat/route.ts lib/openui-cloud/models.ts`
- `pnpm exec prettier --check app/api/chat/route.ts lib/openui-cloud/models.ts`
- `pnpm types:check`
- live compare smoke test: Markdown and OpenUI Cloud both completed successfully
- direct Cloud stream metadata confirmed `openai/gpt-5.4`

### DIFF
--- a/docs/app/api/chat/route.ts
+++ b/docs/app/api/chat/route.ts
@@ -11,7 +11,7 @@ const openUiSystemPrompt = readFileSync(
 
 const markdownSystemPrompt = `You are a helpful assistant. Respond using clear, well-structured GitHub-Flavored Markdown.
 
-Use headings, lists, tables, links, block quotes, and fenced code blocks when they make the response easier to understand. Use the available tools when they are relevant, and incorporate their results into the answer.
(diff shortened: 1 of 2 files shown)


zahlekhan:
Superseded by #828, which includes the compare-only tool scoping correction and points at the current fork branch commit.

PR #835, opened 2026-07-23
feat(react-ui): ambient loading state for AgentInterface + export DotMatrixLoader

ankit-thesys:
Part of the "loaders are just Loading…" cleanup across AgentInterface surfaces (the artifact-renderer side ships separately in the SDK repo).

## What

**`AmbientLoader`** — an AgentInterface-internal loading state: two blurred glow blobs (`blur(64px)`) drift and breathe on opposing 7s/9.5s cycles behind the shared dot-matrix loader, with a **required** contextual label so no call site can render a bare "Loading…" again.

- Theme-token driven: neutral themes render the glow as soft luminance depth; branded themes (a colored accent token, e.g. openui-cloud's blue) tint it automatically.
- `prefers-reduced-motion` freezes the drift; the label is the single `role="status"` announcement (the inner matrix is aria-hidden).
- Internal on purpose — not exported, no compound static.

**Wired into the one full-screen gap**: opening a thread previously hid the entire panel (`visibility: hidden`) with *nothing* shown while messages loaded. That state now renders the loader with "Loading conversation…" — same `isLoadingMessages` flag as the old hiding, so behavior parity is exact. The overlay covers only the thread panel; the sidebar stays interactive.

**`DotMatrixLoader` joins the package index** — artifact renderers are composed by consumers, and their loading states should reuse this loader instead of carrying ports of it.

**Untouched**: every existing DotMatrixLoader site (artifact view page, artifact browser, message loading) — they already do the right thing.

## Test Plan

- [x] `pnpm build` + `pnpm test` (27/27) green; both new classes present in `dist/components/index.css`; `DotMatrixLoader` verified importable from the built index
- [x] Live in the openui-cloud example: thread-open state shows the loader with the brand-blue glow over light theme; sidebar unaffected; settled artifacts still render instantly
- [x] Slow-network sweep (CDP throttling) across browser pages / full-page artifact view confirmed the existing DotMatrixLoader sites behave unchanged

https://claude.ai/code/session_01XECS4NLhz6BNd1TUYbWjfu

### DIFF
--- a/packages/react-ui/src/components/AgentInterface/Thread.tsx
+++ b/packages/react-ui/src/components/AgentInterface/Thread.tsx
@@ -24,11 +24,12 @@ import { Callout } from "../Callout";
 import { DotMatrixLoader } from "../DotMatrixLoader";
 import { IconButton } from "../IconButton";
 import { MarkDownRenderer } from "../MarkDownRenderer";
-import { ResizableSeparator } from "./ResizableSeparator";
-import { UserMessageContent } from "./UserMessageContent";
 import { AgentInterfaceTooltip } from "./_shared/AgentInterfaceTooltip";
 import { GalleryHorizontalEndIcon } from "./_shared/GalleryHorizontalEndIcon";
+import { AmbientLoader } from "./components/AmbientLoader";
+import { ResizableSeparator } from "./ResizableSeparator";
 import { useDetailedViewResize } from "./useDetailedViewResize";
+import { UserMessageContent } from "./UserMessageContent";
 
 export const ThreadContainer = ({
   children,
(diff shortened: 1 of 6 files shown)


packages/react-ui/src/components/AgentInterface/Thread.tsx
```diff
@@ -75,6 +76,15 @@ export const ThreadContainer = ({
...
     >
+      {/* Full-screen loading state while a thread's messages load. The
+          container above hides via `visibility` (keeps layout + scroll state);
+          this overlay opts back in with `visibility: visible`. */}
+      {isLoadingMessages && (
+        <AmbientLoader
+          className="openui-agent-thread-container__loading"
+          label="Loading conversation…"
+        />
+      )}
```
ankit-thesys, line 87: widthin visibility hidden how is this working ?

abhithesys (approved):

PR #871, opened 2026-07-27
Add social metadata to demo pages

zahlekhan:
## Summary

- add a shared page-metadata helper using the refreshed 1800×942 OpenUI social image
- add route-specific canonical, Open Graph, and Twitter card metadata to `/compare`, `/chat`, `/demos`, and `/demo/github`
- preserve the existing titles and descriptions where present, and add matching metadata for the two demo routes that did not have it

## Why

The demo destinations had inconsistent metadata coverage: `/compare` and `/chat` only defined basic page metadata, `/demos` only defined its canonical URL, and `/demo/github` did not define route metadata. As a result, social previews could inherit generic site-level values instead of identifying the destination being shared.

## Impact

Links to each demo now resolve to the correct canonical URL and render a route-specific title and description with a consistent Open Graph/Twitter large-card preview.

## Validation

- `pnpm --filter @openuidev/docs types:check`
- Prettier check for all five changed files
- rendered all four routes locally and verified title, canonical, Open Graph, image dimensions, and Twitter tags

### DIFF
--- a/docs/lib/page-metadata.ts
+++ b/docs/lib/page-metadata.ts
@@ -1,0 +1,45 @@
+import type { Metadata } from "next";
+
+const SOCIAL_IMAGE = "/meta-image.png?v=20260725-1708";
+
+type PageMetadataOptions = {
+  pathname: string;
+  title: string;
+  description: string;
+  imageAlt: string;
+};
+
+export function createPageMetadata({
+  pathname,
+  title,
+  description,
+  imageAlt,
+}: PageMetadataOptions): Metadata {
+  return {
+    title,
+    description,
+    alternates: {
+      canonical: pathname,
+    },
+    openGraph: {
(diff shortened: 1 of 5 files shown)


docs/lib/page-metadata.ts
```diff
@@ -0,0 +1,45 @@
...
+const SOCIAL_IMAGE = "/meta-image.png?v=20260725-1708";
+
+type PageMetadataOptions = {
+  pathname: string;
+  title: string;
+  description: string;
+  imageAlt: string;
+};
+
+export function createPageMetadata({
```
AbhinRustagi, line 12: Can consider adding fallbacks here and then adding this to the root layout which will serve as a default for every new page

AbhinRustagi (approved):

PR #892, opened 2026-07-28
feat: artifactAutoOpen — auto-open artifact detail panels (default on)

ankit-thesys:
## Problem

Artifact detail panels only open on an explicit user click. There is no way for a chat shell to open the panel automatically while an artifact streams in — the behavior the original c1 chat shipped.

## Change

New boolean prop **`artifactAutoOpen`** (**default `true`**, per review) on `ChatProvider`, `AgentInterface`, and the OpenUIChat `withChatProvider` family.

When enabled:

- **The first artifact of a stream opens by itself, once, while it streams.** Further artifacts arriving in the same stream are ignored.
- **A user's close sticks** — nothing re-opens for the rest of the stream (or ever, for that artifact).
- **Edits re-open the panel** (old-c1 parity): the claim is keyed by `id:version`, so an edit's new version gets its own chance and re-points the panel as it streams. Re-registrations of the *same* version stay quiet, so a close still sticks.
- **Thread reloads stay quiet** — historical artifacts register while nothing is running and burn their chance without opening.
- **A new stream's first artifact takes over the panel**: if the panel is already open (user-opened or from a previous stream), it re-points to the new artifact as it starts streaming.

Pass `artifactAutoOpen={false}` for pure click-to-open — the watcher then subscribes to nothing.

## Architecture: registration-driven, entirely in react-headless

`ChatProvider` runs a store-level watcher (`artifactAutoOpenWatcher`) subscribed to the ThreadContext artifact registry. Registration is the trigger — whoever renders an artifact registers it, and that registration is the moment it may present itself. The store layer needs no renderer or parsing knowledge, and react-ui's entire footprint is the prop pass-through (2 lines in `AgentInterface`, 2 in `withChatProvider`).

Design notes:

- **Claim first, ask later**: every first-seen id burns its latch claim whether or not it opens. Historical artifacts can therefore never pop open later, and host remounts / StrictMode re-registrations are invisible.
- **One open per stream**: the watcher tracks `isRunning` transitions; after the stream's first auto-open, later artifacts in the same run are claimed but ignored.
- Both latches live inside the watcher hook — the detailed-view store is untouched by this PR.
- Evaluation is O(registered ids) with zero parsing; already-claimed ids short-circuit.

Artifact sources that bypass the registry (an SDK's chat-library artifacts) are out of scope: a `useArtifactAutoOpen` escape-hatch hook was prototyped and then removed as unused (`a87874dd`; archived in workspace notes). When the chat-shell integration lands it will either register those artifacts in the ThreadContext (watcher covers them) or reintroduce the hook.

## Core files

| File | What it does |
|---|---|
| `react-headless/store/artifactAutoOpenWatcher.ts` | The whole feature in one file. `evaluateRegisteredArtifacts` (one pass: claim once → allowed now → open/take over the panel, at most one per pass) and the `ChatProvider` hook that subscribes it to the artifact registry, holds both latches (per-artifact claims, cleared on thread switch; per-stream open flag, re-armed on each run), and enforces one open per stream. |
| `react-headless/store/ChatProvider.tsx` | Wires the watcher to the three stores — why no other wiring exists anywhere. |
| `react-headless/store/types.ts` | The `artifactAutoOpen?: boolean` prop (default `true`) on `ChatProviderProps`. |
| `react-headless/store/artifactViewId.ts` | `artifactViewId` / `parseArtifactViewId` — single source of truth for the `id:version` view-id format (exported; previously hand-rolled in four places across both packages). |
| `react-ui: AgentInterface.tsx` + `withChatProvider.tsx` | Two lines each — accept the prop, pass it to `ChatProvider`. `ToolActivityRenderer` + `Workspace` now build/parse view ids via the shared helpers instead of hand-rolled strings. |
| Tests (1 suite) | Watcher policy edge cases (close sticks, edits quiet, historical never fires, first-of-pass wins, reset re-arms) and the latch itself. |

## Verification

- react-headless: 122 tests green; typecheck + eslint + prettier clean. react-ui: typecheck, 27 tests, build clean.
- Live e2e on the genui-sdk `testing/openui-cloud` bench (PR head injected, **no prop set** — exercising the default), asserted programmatically and recorded: settled thread load stays quiet; a live generation auto-opens the stream's first artifact mid-stream; a second generation re-points the open panel to its new artifact; an edit re-opens a closed panel on its new version; a mid-stream close sticks through stream end. Labeled recording in the PR comments below.

https://claude.ai/code/session_011nPH2Rn2XGXETgx5Yu5seX


packages/react-headless/src/store/createDetailedViewStore.ts
```diff
@@ -18,7 +18,15 @@ export const createDetailedViewStore = () => {
...
+        set({ activeDetailedViewId: null, _autoOpenedArtifactKeys: new Set() });
+      },
+
+      _autoOpenedArtifactKeys: new Set<string>(),
+      _markAutoOpened: (key) => {
+        const keys = get()._autoOpenedArtifactKeys;
+        if (keys.has(key)) return false;
+        set({ _autoOpenedArtifactKeys: new Set(keys).add(key) });
+        return true;
       },
```
ankit-thesys: Explain this part?

packages/react-headless/src/store/detailedViewTypes.ts
```diff
@@ -35,6 +35,20 @@ export type DetailedViewInternals = {
...
+   * @internal
+   */
+  _autoOpenedArtifactKeys: ReadonlySet<string>;
+  /**
+   * Atomically records an auto-open for `key`. Returns `false` when the key
+   * already fired (the caller must not open again), `true` when this call
+   * claimed it.
+   * @internal
+   */
+  _markAutoOpened: (key: string) => boolean;
```
ankit-thesys: Can you also explain this part?

packages/react-ui/src/components/_shared/tool-renderer/ToolActivityRenderer.tsx
```diff
@@ -134,6 +135,40 @@ export function ToolActivityRenderer<Props>({
...
+      if (!isStreaming) return;
+      if (!dv._markAutoOpened(activity.id)) return;
+    } else {
+      if (openedThisMountRef.current) return;
+    }
+    openedThisMountRef.current = true;
+    dv.setActiveDetailedView(viewId);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [viewMode, isStreaming, isError, parsedIsNull, viewId, activity.id, dvStore]);
+
```
ankit-thesys: Can you tell me why we are doing it here? We can do this in react-headless

Just trying to understand what is happening ?

packages/react-ui/src/components/AgentInterface/AgentInterface.tsx
```diff
@@ -240,6 +241,7 @@ export const AgentInterface: AgentInterfaceComponent = ((props: AgentInterfacePr
         llm={llm}
         artifactRenderers={artifactRenderers}
         artifactCategories={artifactCategories}
+        artifactViewMode={artifactViewMode}
```
ankit-thesys: This looks good.

packages/react-headless/src/store/createDetailedViewStore.ts
ankit-thesys: This is the auto-open latch's check-and-claim, done as one atomic call so there's no gap between "has it fired?" and "mark it fired":

- `keys.has(key)` → already claimed → return `false`, and the caller skips opening. This is what makes a user's mid-stream close stick: if the renderer host remounts during streaming (it does — React StrictMode and streaming re-renders both remount it), the fresh instance asks again, gets `false`, and does NOT re-open over the close.
- Otherwise `new Set(keys).add(key)` — a **copy**, not `keys.add(key)` in place, because zustand subscribers compare by reference; mutating the existing Set would be an invisible state change. Copying keeps the store honest.
- `reset()` (one line up) clears the Set together with the active view on thread switch, so a fresh thread starts with a clean slate.

packages/react-headless/src/store/detailedViewTypes.ts
ankit-thesys: This is the type contract for that latch: `_markAutoOpened(key)` returns a boolean that IS the decision — `true` = this call claimed the key, go ahead and open; `false` = something already auto-opened for this key, do nothing. The caller never reads the Set directly (it's exposed as `ReadonlySet` and underscore-internal, so it's not public API surface).

One thing your comment flushed out: the docstring still said keys are `id:version` from an earlier iteration. The final implementation latches by **tool-call id** — a streamed edit often carries no `version` in its args, so `id:version` collides with the generate's key and would swallow the edit's auto-open (caught in live testing). Tool-call id is unique per generate/edit call and stable across remounts, which is exactly the "once per call" semantics we want. Doc fixed in 4dd3c4c1.

packages/react-ui/src/components/_shared/tool-renderer/ToolActivityRenderer.tsx
ankit-thesys: Fair question — the split is deliberate, and react-headless does own everything it *can* own:

**In react-headless already:** the `artifactViewMode` value (ChatProvider prop → context) and the latch (`_markAutoOpened` on the detailed-view store, reset on thread switch). That's the policy + state.

**Why the trigger effect lives here and not in react-headless:** the open decision needs per-activity *render* context that a renderless package doesn't have —

1. `parsed` / `activity.isError` — a failed or unparseable tool call renders only the fallback card and must never open a panel. Whether a parse succeeded is only known here, where the renderer's parser actually ran.
2. `viewId` — derived in this component, including the `useId()` fallback for early frames before `meta` lands. `useId` is a React render-tree concern; headless can't produce it.
3. `isStreaming` for *this mounted activity* — the reload-safety gate (historical activities mount as `complete`).

ToolActivityRenderer is the one place every artifact renderer is mounted and bound to a viewId — the same place that already runs the two open-*preserving* effects (viewId migration, edit-version follow) this one must cooperate with. If we moved the effect into a react-headless hook, react-ui would still have to call that hook from this exact spot with these exact inputs — same coupling, one extra layer of indirection. So: state/policy headless, render-coupled trigger here.

packages/react-ui/src/components/AgentInterface/AgentInterface.tsx
ankit-thesys: 👍 — and note it's `Omit<ChatProviderProps, "children">` inheritance, so the prop rides the existing type: no separate AgentInterface-side type to keep in sync.

packages/react-ui/src/components/_shared/tool-renderer/ToolActivityRenderer.tsx
ankit-thesys: Update: done your way — 12758b71 moves the whole behavior into react-headless. New exported hook `useArtifactAutoOpen({ viewId, latchKey, isStreaming, enabled })` owns the mode context, the `shouldAutoOpen` predicate, the per-mount latch, and the store-level auto-open latch; `ToolActivityRenderer` shrinks to a single call feeding it the render-derived facts (viewId, tool-call id, streaming, error/parse eligibility) — the only things a renderless package can't know. Bonus: any headless consumer building its own thread UI now gets auto-open from the same hook, and the c1 companion PR deletes its local copy and calls this one (−90 lines there). Predicate matrix test moved into react-headless (118 tests green); full auto-open suite re-verified live on the linked pair after the refactor.

ankit-thesys: Follow-through: `00a81646` moves the invocation too — this file is back to **zero auto-open code**.

ChatProvider now runs a store-level watcher (react-headless `artifactAutoOpenWatcher`): it subscribes to the thread's messages, derives each artifact tool call's status from the store, parses via a shared `runArtifactRenderer` envelope (now also the single definition this renderer uses — the two sides can never disagree on what a parser sees), and opens once per tool call via the existing latch. react-ui's entire diff is the 2+2-line prop pass-through.

Deliberate semantics with the move:
- `open-on-mount` = once per tool call per thread session (loading a thread opens its newest artifact) — a store watcher can't observe mounts.
- Renderers that never yield `meta` (useId-fallback hosts) can't be auto-opened from the store; `useArtifactAutoOpen` stays public for those and for non-tool-call artifact sources (the SDK's chat-library path).
- Cost is latch-bounded: already-opened (or user-closed) calls are skipped before any parsing.

140 tests green (11 new watcher tests); full live matrix re-verified on a linked build: auto-open hands-off, mid-stream close sticks, reload quiet, edit re-opens, presentation, and open-on-mount opening a settled thread.

ankit-thesys:
**Semantics + architecture update (`a4ec3537`), decided in review:** the trigger is now **artifact registration**, and the policy is **"present once"** — an artifact may auto-open exactly one time, when its id first registers, and only into an empty panel.

What changed vs the previous revision:

- **Edits no longer re-open.** An edit shares its generate's artifact id, whose one chance is already spent. An open panel still follows edit versions live (existing follow effect); a closed one stays closed — the user is never interrupted twice by the same artifact.
- **First wins.** An open panel is never stolen — not by a second artifact in the same turn, not by anything.
- **Simpler machinery.** The watcher subscribes to the ThreadContext registry instead of walking messages: no parsing at the store layer, no shared parser envelope needed (`runArtifactRenderer` deleted), and `ToolActivityRenderer` is back to exactly its pre-branch shape. Net **−100 lines**.

Trade-off made deliberately: this drops old c1's "edit re-opens the panel" behavior in exchange for a much simpler mechanism and a calmer UX. Everything re-verified live: auto-open on first registration, close-sticks, **edit stays quiet**, reload quiet, new-id opens, open-on-mount presents a settled thread. 128 headless tests green.

packages/react-headless/src/store/detailedViewTypes.ts
```diff
@@ -35,6 +35,10 @@ export type DetailedViewInternals = {
   _detailedViewPanelNode: HTMLElement | null;
   /** @internal */
   _setDetailedViewPanelNode: (node: HTMLElement | null) => void;
+  /** @internal */
+  _autoOpenedArtifactKeys: ReadonlySet<string>;
+  /** @internal */
+  _markAutoOpened: (key: string) => boolean;
 };
 
 /** Combined detailed-view store type (state + actions + internals). */
```
ankit-thesys, line 40: What is happening here and why are we changing this ?

ankit-thesys: This is the state behind the new `artifactViewMode` prop on ChatProvider. The policy is "present once": an artifact may auto-open the side panel exactly once, when its id first registers — `_autoOpenedArtifactKeys` is the set of keys that already used that chance, and `_markAutoOpened` is the atomic check-and-claim (returns false when the key is already burned, so the caller skips opening). It lives on the detailed-view store rather than component state so the latch survives host remounts and StrictMode double-effects, and `reset()` clears it on thread switch so a new thread starts clean. The underscore + `@internal` marking follows this file's existing convention (`_detailedViewPanelNode`) — nothing here is public API; the barrel only exports the `ArtifactViewMode` type.

packages/react-headless/src/store/ChatProvider.tsx
```diff
@@ -64,14 +67,25 @@ export const ChatProvider: FC<ChatProviderProps> = ({
...
     <ChatContext.Provider value={chatStore}>
       <DetailedViewContext.Provider value={detailedViewStore}>
         <ThreadContextContext.Provider value={threadContextStore}>
           <ArtifactRenderersContext.Provider value={artifactRendererRegistry}>
             <ArtifactStorageContext.Provider value={resolvedStorage.artifact ?? null}>
               <ArtifactCategoriesContext.Provider value={artifactCategories ?? EMPTY_CATEGORIES}>
-                {children}
+                <ArtifactViewModeContext.Provider
+                  value={artifactViewMode ?? DEFAULT_ARTIFACT_VIEW_MODE}
+                >
```
abhithesys: not required

packages/react-headless/src/store/artifactAutoOpenWatcher.ts
```diff
@@ -0,0 +1,53 @@
+import { useEffect } from "react";
```
abhithesys: only this file plus updating props to pass to `useArtifactAutoOpen` is needed.

1. while streaming first artifact should open in that stream.
2. if another artifact comes in same stream, we should ignore it.
3. no need for artifact open on mount. replace it with boolean prop: `autoOpenArtifact`

ankit-thesys: Done in 0b751b99 — exactly this file plus prop plumbing:

1. ✅ While streaming, the stream's **first** artifact opens (registration-driven, while `isRunning`).
2. ✅ Another artifact in the same stream is **ignored** — the watcher latches one open per run (`isRunning` transition resets it), on top of the per-artifact once-latch.
3. ✅ `open-on-mount` removed; the prop is now boolean **`autoOpenArtifact`** (default `false`) on ChatProvider/AgentInterface/withChatProvider. The `ArtifactViewMode` type and mode context are deleted — no new barrel exports at all.

(`useArtifactAutoOpen` itself was already removed earlier as unused — a87874dd.)

120 tests green incl. a new first-of-stream-only case; re-verified live on the genui-sdk bench: settled thread loads stay quiet, live generation auto-opens its first artifact mid-stream, mid-stream close sticks through stream end.

packages/react-headless/src/store/ChatProvider.tsx
ankit-thesys: Removed in 813be897 — the context (and its provider layer here) had no readers left once the escape-hatch hook was deleted. As of 0b751b99 the mode is gone entirely: ChatProvider just passes the boolean `autoOpenArtifact` to the watcher.

packages/react-headless/src/store/artifactAutoOpenWatcher.ts
ankit-thesys: Update per the Linear thread on TH-2454 (`eb58c9d8`): final prop name is **`artifactAutoOpen`** and it now **defaults to `true`** — hosts pass `false` to keep click-to-open. Re-verified live on the bench with no prop set: settled thread quiet, live generation auto-opens its first artifact, mid-stream close sticks.

ankit-thesys:
## E2E demo — final design (`artifactAutoOpen`, default on)

Recorded on the genui-sdk `testing/openui-cloud` bench with **no prop set** (exercising the default), PR head `5644e40c` injected. Real generations, zero artifact clicks; each phase labeled on screen (purple = running, green = passed) and asserted by the driver — the run fails if any behavior is wrong:

artifactAutoOpen final demo

| # | Behavior | Result |
|---|---|---|
| 1 | Loading an **old thread** — nothing streaming → panel stays **closed** | ✓ |
| 2 | **Live generation** — the stream's **first** artifact opens by itself mid-stream | ✓ |
| 3 | **Close mid-stream** — must not re-open for the rest of the stream | ✓ |

(The earlier demo comments showed the since-removed mode-based API and have been deleted.) Full-quality mp4: [gist](https://gist.github.com/ankit-thesys/5634ef8f9f7a0f5b7e1809f4e70cd2a3) · Linear: TH-2454

ankit-thesys:
## Two late behavior changes — recorded and live-verified

Both requested during review of the running build; commits `fd85a15f` and `3d6bde6a`. Driver-asserted runs on the bench (real generations, zero artifact clicks):

### 1. Panel take-over (`fd85a15f`)

A new stream's first artifact **re-points an already-open panel** (previously "first wins" — an open panel was never stolen). In the video: Q2 deck auto-opens → panel left open → EV report requested → panel re-points to the report mid-stream → close mid-stream sticks.

panel take-over demo

### 2. Edit re-open (`3d6bde6a`, old-c1 parity)

The once-claim is keyed by **`id:version`** instead of artifact id, so an edit's new version gets a fresh chance and re-opens even a user-closed panel; same-version re-registrations stay quiet (close sticks per version). In the video: deck auto-opens → user closes → "add one more slide" → panel re-opens on the new version mid-stream → closing during the edit stream sticks.

edit re-open demo

122 unit tests green · PR description updated with both semantics · full-quality mp4s on TH-2454 and in the [gist](https://gist.github.com/ankit-thesys/5634ef8f9f7a0f5b7e1809f4e70cd2a3)

abhithesys (approved):

PR #913, opened 2026-07-31
fix: auto-submit prompt template completions

zahlekhan:
## Summary

- submit a prompt-template completion immediately after composing it with the prefilled prompt stem
- preserve the initial template-chip prefill behavior so users can still edit or submit the stem manually
- update the related API and Storybook descriptions to match the new interaction

## Root cause

The contextual starter handler appended the selected completion to the controlled composer draft and returned focus to the textarea. It never called the thread's message processor, so users had to press Send manually before generation started.

## Impact

Choosing a contextual completion now sends the complete prompt and starts generation in the same click.

## Validation

- `pnpm --filter @openuidev/react-ui test` (27 tests passed)
- `pnpm --filter @openuidev/react-ui typecheck`
- Prettier check on changed files
- `pnpm --filter @openuidev/react-ui lint:check` (0 errors; existing warnings only)
- Storybook interaction verified: template selection followed by completion selection immediately produced the combined user message and loading state

### DIFF
--- a/packages/react-ui/src/components/AgentInterface/ConversationStarter.tsx
+++ b/packages/react-ui/src/components/AgentInterface/ConversationStarter.tsx
@@ -89,7 +89,7 @@ export interface ConversationStarterContainerProps {
   /**
    * Optional click override. When provided, replaces the default
    * send-to-thread behavior (still guarded by `isRunning`). The prefill-chips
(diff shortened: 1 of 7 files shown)


ankit-thesys (approved):
Looks good, pushed one fix to handle the edge case

PR #916, opened 2026-07-31
Group & Display Interleaved Content

AbhinRustagi:
Agent runs that think between tool calls now render as one unit instead of scattered trays and lost text.

- Group consecutive assistant/tool messages into a turn — one "Behind the scenes" timeline per run instead of one per message
- Show the model's thinking as muted text rows between the tool entries, in the order they actually arrived (live runs track arrival order; stored runs use message order)
- Only the final response feeds the Lang renderer — thinking no longer flashes as a fake answer, and the tray stays on "Working…" until the real response starts
- Timeline is height-capped and scrolls internally, smoothly following new rows while streaming (stops when you scroll up, respects reduced motion)
- Form-state saves now round-trip the thinking sections instead of silently deleting them
- New optional messageGroup prop on AssistantMessageComponent — existing custom components are unaffected

No backend, stream-layer, or storage changes.


packages/react-ui/src/components/AgentInterface/Thread.tsx
```diff
@@ -358,6 +417,7 @@ export const Messages = ({
               userMessage={userMessage}
               isStreaming={isRunning && i === lastAssistantIndex}
               isLast={i === lastAssistantIndex}
+              messageGroup={turnByMessageId.get(message.id)}
             />
           </MessageProvider>
         );
```
abhithesys: we should refactor this code. message group + tool call rendering should happen outside of RenderMessage.

packages/react-ui/src/components/AgentInterface/Thread.tsx
```diff
@@ -321,6 +325,190 @@ export const ThreadError = () => {
...
+  className?: string;
+  isRunning: boolean;
+  lastAssistantId: string | null;
+}) => {
+  const assistants = group.filter((m): m is AssistantMessage => m.role === "assistant");
+  const containsLast = !!lastAssistantId && assistants.some((m) => m.id === lastAssistantId);
+  const hasToolCalls = assistants.some((m) => (m.toolCalls?.length ?? 0) > 0);
+
+  if (
+    CustomAssistantMessage &&
```
AbhinRustagi: Remove this

packages/react-ui/src/components/AgentInterface/Thread.tsx
```diff
@@ -321,6 +325,190 @@ export const ThreadError = () => {
...
+        className={className}
+        isRunning={isRunning}
+        lastAssistantId={lastAssistantId}
+      />
+    );
+  }
+
+  return (
+    <>
+      {group.map((message) => (
```
AbhinRustagi: Remove this/ consolidate with if condition

packages/react-ui/src/components/OpenUIChat/GenUIAssistantMessage.tsx
```diff
@@ -144,14 +154,16 @@ export const GenUIAssistantMessage = ({
...
       ))}
-      <Renderer
-        response={openuiCode}
-        library={library}
-        isStreaming={isStreaming}
-        onAction={handleAction}
-        onStateUpdate={handleStateUpdate}
-        initialState={initialState}
-      />
+      {singleResponseStarted && (
```
AbhinRustagi: Check for duplication

abhithesys (approved):

PR #925, opened 2026-08-04
spec: OpenUI 1.0-beta community review draft

Aditya-thesys:
## OpenUI specification, 1.0-beta community review draft

Three documents under `spec/`:

- **overview.md**: what OpenUI is, every feature with examples, how to build with it, and the proposed 1.0-beta extensions (extend, multi-library segments, versioning, registries, data components, constraints).
- **language.md**: the normative language specification: grammar, program model, evaluation, streaming rules, state and forms, data and actions, incremental editing, error recovery, serialization, client checklist.
- **prompt.md**: the LibrarySpec and everything sent to the model: spec documents, prompt generation, signatures and docs, the conversation contract.

Sections 1 to 3 of the overview and everything unmarked in language.md describe shipped behavior; *(proposed)* marks designed but unshipped features, collected in overview section 4 for review.

This is a review draft: feedback welcome on any section, especially the proposed extensions.


spec/overview.md
```diff
@@ -0,0 +1,556 @@
...
+```
+
+Arguments are validated against the schema before the event is dispatched. An unknown action name skips that step with a diagnostic rather than halting the plan, and the remaining steps still run.
+
+### 4.5 Extending a library *(proposed)*
+
+A library derives from another with one method, covering both directions of the common request: use the default library but add your own components, or use it minus the parts your product does not want.
+
+```tsx
+export const library = defaultLibrary.extend({
```
abhithesys, line 485: extending a library means, I have to know each and every component in the library. should we just promote composition instead?

spec/overview.md
```diff
@@ -0,0 +1,556 @@
...
+
+The model tags each fenced block with the library it speaks:
+
+````markdown
+Here is the revenue picture:
+
+```openui-lang library=charts
+root = ChartCard([rev])
+rev = BarChart(labels, [series])
+```
```
abhithesys, line 506: LLM would need to be careful about escaping if it uses backticks. should the fences be something unique like 
```
|>>>> openui-lang lib=charts
Lang
|<<<<
```
 or we expect LLMs to be smart enough?

PR #931, opened 2026-08-05
feat(lang-core): add opt-out postinstall telemetry

zahlekhan:
## Summary

- add a fail-open `postinstall` telemetry hook to `@openuidev/lang-core`
- send one `openui_lang_core_installed` event with native `fetch`
- identify projects using a per-installation salted hash without sending the raw Git origin or path
- include bounded Lang Core, Node, OS, package-manager, CI, and Docker metadata
- support `OPENUI_TELEMETRY_DISABLED=1`, `DO_NOT_TRACK=1`, and stdout debug mode
- add a concise disclosure to the Lang Core README

## Why

This provides a minimal view of Lang Core adoption and installation environments without adding a telemetry SDK dependency or collecting application data.

## User impact

Lang Core installations send one best-effort event when the package manager runs its lifecycle hook. The request has a two-second timeout, cannot fail installation, and can be disabled before project inspection or telemetry state access.

`OPENUI_TELEMETRY_DEBUG=1` prints the payload to stdout without sending it.

## Validation

- Lang Core: 76 tests passed
- typecheck, lint, formatting, and build passed
- publint and package type checks passed
- packed tarball includes the bootstrap and compiled postinstall hook
- clean npm example install emitted exactly one expected `POST /capture/` request
- project ID was hashed and the raw example path was absent
- clean opt-out install emitted no request, state, or notice
- clean debug install printed the payload to stdout and emitted no request


AbhinRustagi:
Could we pull the pure/shared pieces (constants, env checks, identity normalization, CI detection) into something like `src/telemetry/telemetry-shared.ts`?

Referring to the changes in #930

AbhinRustagi (approved):

PR #973, opened 2026-08-13
Start and open CLI apps in interactive mode

zahlekhan:
## Summary

- remove the interactive "Start dev server after install?" prompt and start by default in interactive terminal sessions
- retain `--immediate` as the non-interactive opt-in and `--no-immediate` as the interactive opt-out
- reject `--immediate` and `--no-immediate` when they are passed together
- select an available port with `get-port`, pass it to the dev server through `PORT`, wait for the known localhost URL, and open it

## Why

The extra confirmation interrupted the greenfield create flow even though starting the app is the expected interactive next step. Non-interactive callers still need an explicit opt-in because a long-running server would otherwise change automation behavior.

The CLI constructs the browser URL itself from the numeric port selected by `get-port`; it does not parse framework console output. It waits for an HTTP response before opening the URL. Browser-open failures remain non-fatal and print the URL for manual use.

## Validation

- `pnpm --filter @openuidev/cli build`
- `pnpm --filter @openuidev/cli run ci`
- `git diff --check`
- verified non-interactive default install-and-exit behavior
- verified non-interactive `--immediate` start behavior
- verified interactive default start without the removed prompt
- verified both conflicting-flag orders fail with exit code 1
- fresh npm scaffold installed 649 packages; with port 3000 occupied, `get-port` selected 3001, Next bound to `http://localhost:3001`, returned HTTP 200, and the CLI opened that URL


packages/openui-cli/src/commands/create-app.ts
```diff
@@ -16,6 +17,50 @@ import { resolveAvailableTarget } from "../lib/target-dir";
 import { CliCancelledError, CreateError, telemetry } from "../lib/telemetry";
 import { cliErrorProperties, processErrorProperties } from "../lib/utils";
 
+function isServerReady(url: string): Promise<boolean> {
```
AbhinRustagi: can you move these to lib

AbhinRustagi (dismissed):

AbhinRustagi (approved):

PR #983, opened 2026-08-14
fix: provide theme for assistant-ui renderers

vishxrad:
## Summary

- wrap assistant-ui OpenUI tool renderers in `ThemeProvider` by default
- expose `theme` and `disableThemeProvider` options, matching the existing React integrations
- document CSS layer ordering and theme customization
- add regression coverage for the package-owned provider and host-owned opt-out
- bump `@openuidev/assistant-ui` from `0.0.1` to `0.0.2`

## Root cause

The earlier embedded assistant-ui docs demo owned a custom `OpenUIThemeProvider`. When that demo was removed and the renderer moved into `@openuidev/assistant-ui`, the theme boundary was not carried into the package. The renderer tests continued to wrap `OpenUIPrompt` in `ThemeProvider` manually, which masked the missing production default.

Without the provider, layered OpenUI tokens could follow the system color scheme. A dark system scheme inside a light assistant-ui shell therefore rendered dark-mode text tokens against the light shell background.

## Behavior

The package now initializes light OpenUI theme tokens by default. Consumers can pass any `ThemeProps` through `theme`, or set `disableThemeProvider: true` when the host already owns an OpenUI provider.

## Validation

- `pnpm --filter @openuidev/assistant-ui run ci`
- `pnpm --filter @openuidev/assistant-ui run check:publint`
- `pnpm --filter @openuidev/assistant-ui run check:attw`
- reproduced and visually verified in an assistant-ui shell

### DIFF
--- a/packages/assistant-ui/README.md
+++ b/packages/assistant-ui/README.md
@@ -11,8 +11,15 @@ pnpm add @openuidev/assistant-ui @assistant-ui/react @openuidev/react-ui @openui
 Import the OpenUI styles once in your app:
 
 ```css
+@layer theme, base, openui, components, utilities;
+
 @import "@openuidev/react-ui/layered/styles/index.css";
 ```
+
+The layer declaration keeps OpenUI above Tailwind's reset and below application
+components and utilities. The tool renderers include OpenUI's light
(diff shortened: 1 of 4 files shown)


ankit-thesys (approved):
LG

PR #986, opened 2026-08-14
fix react-ui invalid chart geometry

vishxrad:
## Summary

- skip LineInBarShape rendering while Recharts supplies non-finite responsive-layout geometry
- render normally again once a later layout pass supplies valid dimensions
- add regression coverage for valid SVG output and non-finite x, y, width, and height values

## Root cause

Recharts can briefly provide NaN dimensions while measuring a responsive bar chart, including charts in containers that have not finished layout. LineInBarShape used those values as SVG path and line coordinates, producing invalid DOM attributes and React warnings.

The custom shape now validates all DOM-bound geometry at its render boundary and returns null only for the invalid pass. It remains stateless and renders normally when Recharts provides finite geometry.

## Validation

- pnpm --filter @openuidev/react-ui run ci (39 tests)
- pnpm --filter @openuidev/react-ui run typecheck
- pnpm --filter @openuidev/react-ui run build

Extracted from https://github.com/thesysdev/openui/pull/985 to keep the assistant-ui package change focused.

### DIFF
--- a/packages/react-ui/src/components/Charts/shared/LineInBarShape/LineInBarShape.test.tsx
+++ b/packages/react-ui/src/components/Charts/shared/LineInBarShape/LineInBarShape.test.tsx
@@ -1,0 +1,32 @@
+import { renderToStaticMarkup } from "react-dom/server";
+import { describe, expect, it } from "vitest";
+import { LineInBarShape } from "./LineInBarShape";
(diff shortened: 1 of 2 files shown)


ankit-thesys (approved):
Looks good

PR #1023, opened 2026-08-19
Add Recipe Remix Cloud quickstart

zahlekhan:
## Summary

- replace the finance quickstart with a Cloud-focused Recipe Remix tutorial
- add recipe starter prompts, a pantry function tool, and an interactive custom component
- use Cloud web and image search while preserving the built-in library and artifact renderers
- generate a serializable component spec so the API route does not import React renderers

## Why

The quickstart now ends with a useful agent that demonstrates both ways users extend a Cloud starter: adding a server-side tool and adding a Generative UI component. The generated spec keeps the React component library on the client side and avoids Next.js React Server Component runtime failures.

## Validation

- generated the spec with `@openuidev/cli@latest` in a fresh published Cloud scaffold
- `npm run typecheck` in the scaffold
- focused ESLint checks for the component, API route, and Next config
- `npm run build` in the scaffold with Next.js 16.3.1
- Prettier check for the quickstart MDX
- verified the final page in the local docs preview


AbhinRustagi:
Why don't we update the existing template? 

Also this code block is scrollable in both axes, can we word-wrap/prettify and make it vertical scroll only?
Screenshot 2026-08-26 at 15 16 39

zahlekhan:
Addressed the code-block feedback in `083838db`: the Recipe Remix component block now wraps long lines, hides horizontal overflow, and keeps vertical scrolling.

Validated with:
- `pnpm types:check`
- `pnpm exec prettier --check content/docs/openui-lang/quickstart.mdx`

Quickstart recording:

Recipe Remix quickstart

AbhinRustagi (approved):

PR #1065, opened 2026-08-24
fix(observability-cloud): report the real SDK version on the wire (0.0.3)

ankit-thesys:
## Problem

`SDK_VERSION` in `src/core/wire.ts` was never bumped for the 0.0.2 release, so it still read `"0.0.1"`:

```ts
/** Must be kept in sync with packages/observability-cloud/package.json on release. */
export const SDK_VERSION = "0.0.1";   // package.json: 0.0.2
```

That constant is what every envelope reports:

```json
{"v":1,"sdk":{"name":"observability-cloud","version":"0.0.1"},...}
```

So **every 0.0.2 client identifies itself as 0.0.1 on the wire.** Ingest cannot distinguish the two releases, which makes any server-side client-version analysis wrong — including the question you most want answered after shipping a fix: *how many clients have actually picked it up?*

Confirmed against a real envelope from a 0.0.2 install.

## Change

Releases **0.0.3** with the constant corrected:

- `package.json` — `0.0.2` → `0.0.3`
- `src/core/wire.ts` — `SDK_VERSION` `"0.0.1"` → `"0.0.3"`

Plus a build-time guard so it cannot drift again — see the comment below.

No behaviour change beyond the reported version.

## Note: the invariant was already covered by a test

`src/core/wire.test.ts` already asserts it:

```ts
expect(SDK_VERSION).toBe(version);   // version read from package.json
```

On current `main` that test fails:

```
FAIL  src/core/wire.test.ts > SDK_VERSION > matches package.json version
AssertionError: expected '0.0.1' to be '0.0.2'
```

So no new test is needed — this PR makes the existing one pass. Might be worth a look at whether this package's tests run in CI.

## Verification

Baseline on `main`: `1 failed | 46 passed` — the sole failure being that guard.

With this change:

- `pnpm test` — **47 passed (10 files)**
- `pnpm typecheck` — clean
- `pnpm run lint:check` / `format:check` — clean
- `pnpm run build` → built dist emits `SDK_VERSION = "0.0.3"`

### DIFF
--- a/packages/observability-cloud/package.json
+++ b/packages/observability-cloud/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openuidev/observability-cloud",
-  "version": "0.0.2",
+  "version": "0.0.3",
   "description": "Cloud sink for @openuidev/observability: batches OpenUI events and ships them to Thesys ingest",
   "license": "MIT",
   "type": "module",
--- a/packages/observability-cloud/package.json
+++ b/packages/observability-cloud/package.json
@@ -26,7 +26,8 @@
   },
   "scripts": {
     "test": "vitest run --passWithNoTests",
-    "build": "tsdown",
+    "check:version-sync": "node scripts/check-version-sync.mjs",
+    "build": "pnpm run check:version-sync && tsdown",
     "watch": "tsdown --watch",
     "typecheck": "tsc --noEmit",
     "lint:check": "eslint ./src",
--- a/packages/observability-cloud/scripts/check-version-sync.mjs
+++ b/packages/observability-cloud/scripts/check-version-sync.mjs
@@ -1,0 +1,58 @@
+#!/usr/bin/env node
+/**
+ * Fails the build when `SDK_VERSION` in src/core/wire.ts has drifted from the
+ * version in package.json.
+ *
+ * Why this exists: `SDK_VERSION` is what every envelope reports as
+ * `sdk.version`, and it is maintained by hand. It was missed on the 0.0.2
+ * release, so every 0.0.2 client identified itself as 0.0.1 on the wire and
+ * ingest could not tell the two releases apart.
+ *
+ * There is already a unit test asserting this (src/core/wire.test.ts), but a
+ * release can be cut without the tests passing — that is exactly how 0.0.2
+ * shipped. This runs as part of `build`, which `prepare` invokes, so it also
+ * guards `npm publish`.
+ *
+ * Deliberately wired into the `build` script rather than named `prebuild`:
+ * pnpm ships with `enable-pre-post-scripts=false` by default, so a `prebuild`
+ * script would silently never run.
+ */
+
+import { readFileSync } from "node:fs";
+import { dirname, join } from "node:path";
+import { fileURLToPath } from "node:url";
+
+const packageRoot = join(dirname(fileURLToPath(import.meta.url)), "..");
+const packageJsonPath = join(packageRoot, "package.json");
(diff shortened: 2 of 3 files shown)


ankit-thesys:
Pushed a second commit adding a **build-time guard** so this cannot drift again.

`packages/observability-cloud/scripts/check-version-sync.mjs` compares `SDK_VERSION` against `package.json` and exits non-zero on a mismatch:

```
  version-sync check failed

  SDK_VERSION is "0.0.1" but package.json is "0.0.3".

  SDK_VERSION is reported on every envelope as sdk.version, so a mismatch
  makes ingest unable to tell releases apart.

  Fix: set SDK_VERSION to "0.0.3" in src/core/wire.ts
```

**Two deliberate choices worth reviewing:**

**It is chained into `build`, not named `prebuild`.** pnpm ships with `enable-pre-post-scripts=false` by default and this repo has no `.npmrc` overriding it — so a script named `prebuild` would silently never run. `"build": "pnpm run check:version-sync && tsdown"` always runs, and because `prepare` invokes `build`, it guards `npm publish` as well.

**It duplicates what the unit test already asserts, on purpose.** `wire.test.ts` had this invariant covered and still let 0.0.2 ship wrong, because a release can be cut without the tests passing. A build cannot be cut without this. The test tells you during development; the build stops you at the door.

`scripts/` is not in the package `files` array, so nothing extra ships in the tarball.

Verified both directions: passes in sync, and reverting `SDK_VERSION` to `"0.0.1"` fails with exit 1 *before* `tsdown` runs. Full package still green — 47 tests, typecheck, lint, format.

**Alternative if you'd prefer it:** derive `SDK_VERSION` from `package.json` at build time (tsdown `define`, or a generated file) so the constant cannot be wrong by construction. That removes the failure mode entirely rather than detecting it, but it is a bigger change to how the package builds — happy to swap if you'd rather go that way.

packages/observability-cloud/scripts/check-version-sync.mjs
```diff
@@ -0,0 +1,58 @@
...
+ * There is already a unit test asserting this (src/core/wire.test.ts), but a
+ * release can be cut without the tests passing — that is exactly how 0.0.2
+ * shipped. This runs as part of `build`, which `prepare` invokes, so it also
+ * guards `npm publish`.
+ *
+ * Deliberately wired into the `build` script rather than named `prebuild`:
+ * pnpm ships with `enable-pre-post-scripts=false` by default, so a `prebuild`
+ * script would silently never run.
+ */
+
```
AbhinRustagi: don't need a big comment

packages/observability-cloud/package.json
```diff
@@ -26,7 +26,8 @@
   },
   "scripts": {
     "test": "vitest run --passWithNoTests",
-    "build": "tsdown",
+    "check:version-sync": "node scripts/check-version-sync.mjs",
+    "build": "pnpm run check:version-sync && tsdown",
```
AbhinRustagi, line 30: check if there is a prebuild script convention?

packages/observability-cloud/scripts/check-version-sync.mjs
ankit-thesys: removed it.

packages/observability-cloud/package.json
ankit-thesys: There is none, but we have followed similar script checks in react-ui and react-headless in post-build.

ankit-thesys:
already fixed by @abhithesys

PR #1068, opened 2026-08-25
fix(react-ui): don't submit on Enter while an IME composition is open

### LINKED ISSUE #1045: Windows voice to text leaves text in the prompt composer
## Tracking

This report covers two submission paths, tracked as separate subissues:

- #1226 — Enter/IME handling, addressed by PR #1068.
- #1227 — Investigate the reported Send-button race during Windows Voice Typing.

Keep this parent issue open until both paths have been resolved or explicitly dispositioned.

---

**Describe the bug**
I am a heavy user of the voice typing that is part Windows. (Win+h). Under certain circumstances if you use Windows Voice typing to enter text in the prompt area If you hit the enter key or click the send button, the text remains in the prompt without being cleared.

**To Reproduce**
1. Go to https://www.openui.com/chat?form_factor=desktop
2. Click in the prompt area
3. Hit Win+h and dictate some text
4. Hit the enter key or click the send button

Note that the issue appears to be some sort of a race issue. You may need to hit enter pretty quickly.

**Expected behavior**
The prompt area should be clear.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser Edge  151.0.4129.101, Chrome 151.0.7922.174

**Additional context**
I was able to resolve the issue by creating a custom composer which is mainly a copy of Composer.tsx with a few modifications clearly documented below.

```ts
// Windows Voice Typing (dictation) commits recognized speech into the field through an IME
// composition session that, per the report this fix is based on, stays open for as long as
// dictation is active — even across a real, physical Enter keypress typed by hand (not a spoken
// "Enter" command) once the user believes a phrase is done. The library's built-in composer
// (@openuidev/react-ui's AgentInterface/components/Composer.tsx) has no composition awareness at
// all: its onKeyDown treats *any* Enter as submit, including one that lands while
// `event.isComposing` is still true. So handleSubmit fires and clears the box on whatever text
// has streamed in so far, and then the still-open composition session finalizes a moment later
// and fires one more onChange with the (re-)committed dictated text, which repopulates the
// now-empty field — reproduced as "clears, then the spoken text reappears." (With DevTools open,
// event timing shifts enough that the race doesn't happen and everything looks fine — that was
// the first symptom reported.) assistant-ui's own composer (ComposerInput.js) guards against
// exactly this by bailing out of Enter-to-submit whenever `e.nativeEvent.isComposing` (or a
// compositionstart/end-tracked ref, since isComposing's timing relative to keydown isn't fully
// consistent across browsers) is true — the mirrored guard below. This is a full reimplementation
// via AgentInterface.Composer's documented "Mode C" (children replace the composer entirely)
// rather than a fix inside the library, since there's no prop-level hook into the built-in
// composer's internals. See FRAMEWORK_ISSUES.md's OpenUI section.
//
// Trade-off inherent to this fix, not a bug: while the composition session is still open, Enter
// no longer submits — it's left to the browser/IME to finalize first, same as any other IME-based
// input. If Voice Typing keeps composition open across the whole dictation session, the first
// physical Enter after dictating may just close out composition without sending, and a second
// Enter is needed to actually submit. That's standard behavior for IME input generally, not
// something to chase further — doing so is what caused the original race.
function OpenUiComposer() {
  const { processMessage, cancelMessage, isRunning, isLoadingMessages } = useThread();
  const [textContent, setTextContent] = useState("");
  const inputRef = useRef<HTMLTextAreaElement>(null);
  // FIX (not in upstream Composer.tsx): ref-tracked fallback for `e.nativeEvent.isComposing`,
  // read in onKeyDown below — see the file-level comment on why isComposing's timing isn't trusted alone.
  const isComposingRef = useRef(false);
  const [hasOverflowTop, setHasOverflowTop] = useState(false);
  const [hasOverflowBottom, setHasOverflowBottom] = useState(false);

  const updateOverflow = useCallback(() => {
    const input = inputRef.current;
    if (!input) return;
    const maxScrollTop = input.scrollHeight - input.clientHeight;
    setHasOverflowTop(maxScrollTop > 0 && input.scrollTop > 0);
    setHasOverflowBottom(maxScrollTop > 0 && input.scrollTop < maxScrollTop - 1);
  }, []);

  const handleSubmit = () => {
    if (!textContent.trim() || isRunning || isLoadingMessages) return;
    processMessage({ role: "user", content: textContent });
    setTextContent("");
    // FIX (not in upstream Composer.tsx): the Send button's onClick below calls handleSubmit
    // directly, with no composing check — unlike the Enter-key path, this one can still fire mid-
    // dictation. Forcing the DOM value empty (not just React state) narrows the same race for that
    // path: without it, e.currentTarget.value in onCompositionEnd could still read old-plus-newly-
    // committed text once the session finalizes. Best-effort only — setting .value mid-composition
    // isn't guaranteed to reset the IME consistently across browsers.
    if (inputRef.current) inputRef.current.value = "";
  };

  useLayoutEffect(() => {
    const input = inputRef.current;
    if (!input) return;
    input.style.height = "0px";
    input.style.height = `${Math.max(input.scrollHeight, 24)}px`;
    updateOverflow();
  }, [textContent, updateOverflow]);

  return (
    <div
      className="openui-agent-thread-composer"
      onClick={(e) => {
        if (!(e.target as HTMLElement).closest("button, a, [role='button']")) inputRef.current?.focus();
      }}
    >
      <div
        className="openui-agent-thread-composer__input-wrapper"
        data-overflow-top={hasOverflowTop || undefined}
        data-overflow-bottom={hasOverflowBottom || undefined}
      >
        <textarea
          ref={inputRef}
          value={textContent}
          autoFocus
          onChange={(e) => setTextContent(e.target.value)}
          onScroll={updateOverflow}
          // FIX (not in upstream Composer.tsx): upstream has no onCompositionStart/onCompositionEnd
          // handlers at all — these are the composition-tracking half of the guard.
          onCompositionStart={() => {
            isComposingRef.current = true;
          }}
          onCompositionEnd={(e) => {
            isComposingRef.current = false;
            // Syncs the IME's finalized text: compositionend can commit text without a preceding
            // onChange in some browsers, so relying on onChange alone would drop the last commit.
            setTextContent(e.currentTarget.value);
          }}
          className="openui-agent-thread-composer__input"
          placeholder="Type your query here"
          rows={1}
          onKeyDown={(e) => {
            if (e.key === "Enter" && !e.shiftKey) {
              // FIX (not in upstream Composer.tsx): upstream's onKeyDown goes straight to
              // preventDefault()+handleSubmit() here with no composition check at all — this bails
              // out instead, leaving Enter to the browser/IME while dictation is still composing.
              if (e.nativeEvent.isComposing || isComposingRef.current) return;
              e.preventDefault();
              handleSubmit();
            }
          }}
        />
        <div className="openui-agent-thread-composer__action-bar">
          <IconButton
            onClick={isRunning ? cancelMessage : handleSubmit}
            icon={isRunning ? <Square size="1em" fill="currentColor" /> : <ArrowUp size="1em" />}
            size="extra-small"
            variant="primary"
            aria-label={isRunning ? "Cancel message" : "Send message"}
            className="openui-agent-thread-composer__submit-button"
          />
        </div>
      </div>
    </div>
  );
}
```

iambharathpadhu:
## What

Fixes #1045.

Both built-in composers in `@openuidev/react-ui` submit on **any** `Enter` keydown, with no awareness of IME composition state — there is currently no composition guard anywhere in the package (`grep -rn "isComposing" packages/react-ui/src` returns nothing).

Windows Voice Typing holds an IME composition session open across dictation. A physical `Enter` therefore submits and clears the textarea, the still-open composition finalizes a moment later and fires one more `onChange`, and the dictated text reappears in the box the user just emptied. That is the race in the issue.

The blast radius is wider than the report suggests: **every IME-based input hits the same path**. For Japanese, Chinese and Korean input, `Enter` is the key that *confirms* a conversion candidate — today that keypress sends a half-converted phrase instead of committing it.

The issue reporter worked around this in their own app via `AgentInterface.Composer`'s Mode C, patching one composer. This fixes it in the library, for both.

## Changes

- Add `shouldSubmitOnEnter()` in `AgentInterface/_shared/utils/composerKeyboard.ts` — a pure predicate that returns `false` while a composition is open.
  - Checks `nativeEvent.isComposing`, **plus** the `keyCode === 229` sentinel, since Safari and older Chromium dispatch the composition-consumed `keydown` before `isComposing` is set.
  - Takes a narrow structural `ComposerKeyDownEvent` that React's `KeyboardEvent<HTMLTextAreaElement>` satisfies, so call sites pass the synthetic event straight through and tests build a plain object — no casts, no DOM.
- Route both composers through it, so the condition has one source of truth instead of two copies:
  - `AgentInterface/components/Composer.tsx`
  - `AgentInterface/components/DesktopWelcomeComposer.tsx`
- Add `__tests__/composerKeyboard.test.ts` covering plain `Enter`, `Shift+Enter`, non-`Enter` keys, `isComposing`, the `229` sentinel, and the `Enter` that follows a finished composition.

### Behavior change

While a composition is open, `Enter` no longer sends — it closes the composition, and the next `Enter` sends. This is standard IME behavior, matches what `assistant-ui`'s own composer does, and is precisely what removes the race. Non-IME typing is unaffected: `Enter` sends, `Shift+Enter` inserts a newline, exactly as before.

### On the approach

I went with a shared pure helper rather than a `compositionstart`/`compositionend` ref inside each component for two reasons: it keeps the two composers from drifting apart again, and it stays testable under the package's existing setup. `react-ui` has no jsdom today — every test here is SSR or pure — so a component-level interaction test would mean adding a DOM environment, a larger diff than the fix itself. Happy to inline the guard in both components instead if you'd prefer that.

## Test Plan

- [x] Verified locally

```
pnpm --filter @openuidev/react-ui run ci
```

- `lint:check` — 0 errors (27 pre-existing `react-hooks/exhaustive-deps` warnings in `genui-lib`/`hooks`, none from these files)
- `format:check` — clean
- `test` — **40 passed across 8 files** (6 new)

`pnpm --filter @openuidev/react-ui typecheck` reports 73 errors, but the count is byte-identical on a clean checkout of `main` — all in `genui-lib`, none in `AgentInterface`. It is also not part of the package's `ci` script, so I left it alone rather than widening this PR.

## Checklist

- [x] I updated docs/README when needed — no public API change; the helper is internal to `AgentInterface`
- [x] I considered backwards compatibility — no exported surface changes; the only behavioral difference is the IME case described above

### DIFF
--- a/packages/react-ui/src/components/AgentInterface/__tests__/composerKeyboard.test.ts
+++ b/packages/react-ui/src/components/AgentInterface/__tests__/composerKeyboard.test.ts
@@ -1,0 +1,37 @@
+import { describe, expect, it } from "vitest";
+import { ComposerKeyDownEvent, shouldSubmitOnEnter } from "../_shared/utils/composerKeyboard";
+
+const keyDown = (overrides: Partial<ComposerKeyDownEvent> = {}): ComposerKeyDownEvent => ({
+  key: "Enter",
+  shiftKey: false,
+  keyCode: 13,
+  nativeEvent: { isComposing: false },
+  ...overrides,
+});
+
+describe("shouldSubmitOnEnter", () => {
+  it("submits on a plain Enter", () => {
+    expect(shouldSubmitOnEnter(keyDown())).toBe(true);
+  });
+
+  it("does not submit on Shift+Enter — that inserts a newline", () => {
+    expect(shouldSubmitOnEnter(keyDown({ shiftKey: true }))).toBe(false);
+  });
+
+  it("ignores keys other than Enter", () => {
+    expect(shouldSubmitOnEnter(keyDown({ key: "a", keyCode: 65 }))).toBe(false);
+  });
+
+  it("does not submit while an IME composition is open (isComposing)", () => {
+    expect(shouldSubmitOnEnter(keyDown({ nativeEvent: { isComposing: true } }))).toBe(false);
+  });
+
+  it("does not submit when the browser reports the IME sentinel keyCode 229", () => {
+    // Safari and older Chromium leave isComposing unset on this keydown.
+    expect(shouldSubmitOnEnter(keyDown({ keyCode: 229 }))).toBe(false);
+  });
+
+  it("submits on the Enter that follows a finished composition", () => {
+    expect(shouldSubmitOnEnter(keyDown({ nativeEvent: { isComposing: false } }))).toBe(true);
+  });
+});
--- a/packages/react-ui/src/components/AgentInterface/_shared/utils/composerKeyboard.ts
+++ b/packages/react-ui/src/components/AgentInterface/_shared/utils/composerKeyboard.ts
@@ -1,0 +1,40 @@
+/**
+ * The subset of a `keydown` event the composers need to decide whether `Enter`
+ * submits. React's `KeyboardEvent<HTMLTextAreaElement>` satisfies it
+ * structurally, so call sites pass the synthetic event straight through and
+ * tests can build a plain object.
+ */
+export interface ComposerKeyDownEvent {
+  key: string;
+  shiftKey: boolean;
+  keyCode: number;
+  nativeEvent: { isComposing: boolean };
+}
+
+/**
+ * Safari and older Chromium report `keyCode` 229 for a keydown consumed by an
+ * open IME composition, and do not always have `isComposing` set on that same
+ * event. Checking both covers each browser's timing.
+ */
+const IME_KEY_CODE = 229;
+
+/**
+ * Whether an `Enter` keydown in a composer textarea should send the draft.
+ *
+ * `Enter` sends and `Shift+Enter` inserts a newline — except while an IME
+ * composition is open, where `Enter` belongs to the IME (it commits the
+ * conversion candidate) and must not reach the composer.
+ *
+ * Without this guard, Windows Voice Typing holds a composition session open
+ * across dictation: a physical `Enter` sends and clears the textarea, the
+ * composition then finalizes and fires one more `onChange`, and the dictated
+ * text reappears in the box the user just emptied. Every CJK IME hits the same
+ * path and sends a half-converted phrase instead of committing it.
+ *
+ * Consequence, shared with every IME-aware composer: the `Enter` that closes a
+ * composition does not send. The next one does.
+ */
+export const shouldSubmitOnEnter = (event: ComposerKeyDownEvent): boolean => {
+  if (event.key !== "Enter" || event.shiftKey) return false;
+  return !event.nativeEvent.isComposing && event.keyCode !== IME_KEY_CODE;
+};
--- a/packages/react-ui/src/components/AgentInterface/components/Composer.tsx
+++ b/packages/react-ui/src/components/AgentInterface/components/Composer.tsx
@@ -6,6 +6,7 @@ import { useLayoutContext } from "../../../context/LayoutContext";
 import { useAutoFocus } from "../../../hooks/useAutoFocus";
 import { useComposerState } from "../../../hooks/useComposerState";
 import { IconButton } from "../../IconButton";
+import { shouldSubmitOnEnter } from "../_shared/utils/composerKeyboard";
 
 export interface ComposerProps {
   className?: string;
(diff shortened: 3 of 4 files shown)


vishxrad (approved):
LGTM. Thanks for your contribution :)

PR #1077, opened 2026-08-26
Add the Generative UI benchmark page

rahuldoval:
## Summary
- add the new Generative UI benchmark page and 30-model OpenUI comparison
- keep the six-model format comparison as the second chart view
- expose exact benchmark values as semantic HTML, JSON, CSV, Markdown, and JSON Schema
- make both top chart tabs readable to fetch-only crawlers and operable by browser agents
- add agent-answer contracts and crawler/publication checks

## Verification
- `pnpm types:check`
- `pnpm check:benchmark-agents` (20/20 answer contracts)
- `pnpm audit:benchmark-publication`
- `pnpm build`
- interactive browser-agent test for tab roles, selection, panels, and chart labels

## Evidence status
The six-model format comparison links published raw results and outputs. The 30-model board is explicitly marked summary-only until row-level raw generations and pricing evidence are published.


rahuldoval:
## External agent-readability test

The Vercel preview built successfully, but anonymous requests are redirected to Vercel SSO, so it is not a valid general-agent test URL while Deployment Protection is enabled.

A production build was exposed through a temporary Cloudflare quick tunnel and tested in three ways:

- fetch-only crawler audit as Googlebot, Bingbot, and OAI-SearchBot: passed
- semantic/data consistency and answer contracts: 20/20 passed
- browser agent: discovered both tabs, clicked Format comparison, observed the selected state and controlled panel change, and read the chart label

An independent agent given only the public URL correctly reported:

- Model comparison = 30 OpenUI models; Format comparison = six models across OpenUI, A2UI, and json-render
- Grok 4.6 and GPT-5.6 Sol tie at 99.5% OpenUI validity
- OpenUI format summary = 96.5% structural validity and 99.9% render success
- self-hosted/null cost means cost is not comparable, not zero

The agent could operate and read the HTML page even though its browser client blocked direct opening of the JSON/CSV/Markdown endpoints. This validates keeping both exact tab datasets in the server-rendered HTML. The Cloudflare quick-tunnel URL is ephemeral and should not be treated as a review deployment.

PR #1095, opened 2026-08-27
Restructure the home page around the three OpenUI products

pd-builds:
Work in progress, opened so it can be picked up by someone else. Not ready to merge: see **Outstanding** below.

## What changed

**One shared `ProductSection`, rendered three times.** Lang on the page background, Gateway and Observability on the black Cloud band. They differ only in copy and a `tone` prop, so they read as three of the same thing. This replaces the stack diagram and the use-cases stepper, both removed from the home page.

**The Cloud band is now just the products.** Its heading, feature grid and capability ticker are gone; the contact CTA moved to the top and took the section-title scale ("Upgrade your Generative UI agent.").

**The feature grid section keeps only its benchmark header and the compatibility band**, in that order, via two new opt-in props (`compatFirst`, `showHeaderSeparator`). The four other pages that use this component are unaffected: all pass `showCompat={false}` and keep the defaults.

**New closing FAQ**, built on `<details>` rather than a JS accordion so every answer stays in the markup for crawlers and coding agents, and it works with no JavaScript.

**The agent picker copies a setup prompt**, not a CLI command. The per-agent `--agent-name` suffix is removed so the prompt is copied verbatim; both pickers (header and hero) share one exported `AGENT_SETUP_PROMPT`.

## Outstanding

- **Artwork is missing.** Each band's stage renders empty at the right aspect ratio. The brief for all three images is recorded at the top of `products.ts`.
- **Gateway's "View docs" is a placeholder**, pointing at the Cloud docs, since Gateway has no docs of its own yet. Flagged in the file.
- **The home page no longer links to `/cloud/observability`.** The latest copy dropped each band's primary CTA, leaving only "View docs". The nav still links there.
- **"3 Million+ downloads" is hardcoded** beside the live star count. Worth confirming against the real npm figure.

## Verification

`next build` passes. Layout, spacing, copy, theming and the clipboard payload were checked in the browser against both light and dark themes.


rahuldoval (approved):
Final website copy and asset QA completed; all required checks passed.

PR #1132, opened 2026-09-08
Remove CLOUD_EMBED env var

AbhinRustagi:
- Hard writes the OpenUI Cloud API URL since it's not configurable or replaceable when using Cloud


vishxrad (approved):
LGTM

PR #1135, opened 2026-09-08
Fix self-hosted Vercel Eve CLI template build

AbhinRustagi:
## Summary
- Install `@openuidev/react-lang` and `zustand` in the self-hosted Vercel Eve overlay so npm actually gets those OpenUI peers under `legacy-peer-deps`.
- Allow Eve's `.ts` imports during typecheck, and turn off Turbopack minify to match the working Eve example.

## Test plan
- [x] `npx @openuidev/cli@latest create --name demo --template openui-self-hosted --backend-framework vercel-eve --no-interactive --no-skill`
- [x] `OPENAI_API_KEY=sk-test npm run build` in the generated app


i-subham23 (approved):
Looks Good

PR #1140, opened 2026-09-09
fix(lang-core): use last complete definition for duplicate IDs in streaming parser

### LINKED ISSUE #1127: createStreamParser() and parse() disagree on redefined statement IDs
parse() and createStreamParser() produce different final results for the same completed OpenUI Lang program when a statement ID is defined more than once.

The batch parser behaves as last definition wins, while the streaming parser behaves as first definition wins.

Minimal reproduction

const text = `root = Stack([a])
a = Title("x")
a = Title("y")
`;

const batch = parse(text, schema);
// root -> Title("y")

const stream = createStreamParser(schema);
stream.push(text);
const streamed = stream.getResult();
// root -> Title("x")

Actual behavior

For the same completed input:

parse(text)                  -> Title("y")
createStreamParser().push()  -> Title("x")

I also tested multiple delivery schedules:

the whole input in a single push()

line-by-line

character-by-character

multiple two-chunk splits

The streaming parser consistently returns "x" in all of these cases, while parse() consistently returns "y".

So this does not appear to be a chunk-boundary race. It looks like a systematic difference in duplicate/redefined statement semantics between the batch and streaming parsers.

Likely cause

The batch parser builds its statement map using normal Map.set() behavior, so a later statement with the same ID replaces the earlier one.

The streaming parser caches completed statements and skips a later completed statement when the ID is already present, effectively making that path first-definition-wins.

In particular, the completed-statement cache together with the existing guard around an already-present statement ID appears to preserve the first definition.

Expected behavior / question

I think parse() and createStreamParser() should have the same final semantics for the same completed input.

I'm not sure which duplicate-ID behavior is intended:

last definition wins

first definition wins

duplicate statement IDs are invalid and should produce a parser error

Happy to contribute the corresponding fix and regression coverage once the intended behavior is confirmed.

Additional context

I found this while building a streaming-invariance test harness for createStreamParser().

The harness currently checks valid OpenUI Lang programs across:

exhaustive single-cut partitions

exhaustive two-cut partitions

seeded multi-cut partitions

trailing-newline variants

no-prefix-crash behavior

error convergence

determinism

Those tests are currently green. Duplicate statement IDs were deliberately excluded because their intended semantics were unclear; testing that excluded case exposed this batch-vs-stream divergence.

AbhinRustagi:
## What

Fix streaming parsing of redefined statement IDs so the last complete definition replaces the previous one, matching the existing non-streaming behavior.

Closes #1127.

```text
root = Stack([a])
a = Title("x")
a = Title("y")
```

Previously, `parse()` rendered `"y"` while the streaming parser rendered `"x"`. Both now render `"y"`, with or without a trailing newline.

## Changes and rationale

- **Preserve trailing newlines only for streaming.** The streaming scanner uses newlines outside strings and brackets as statement boundaries. Trimming the final newline hid that boundary and kept the final statement in the pending path. Preserving it lets the scanner commit the statement to its cache immediately. The boundary newline is not part of the rendered text; newlines inside closed strings were already preserved.
- **Allow complete pending replacements.** The model may omit the final newline. A pending statement that does not need quotes or brackets supplied by `autoClose()` can now overwrite an existing ID. This is what lets a complete final replacement take effect without a newline.
- **Protect completed values during partial replacements.** If `autoClose()` must supply missing quotes or brackets, the pending statement cannot overwrite a completed ID. For example, streaming `a = Title("y` keeps the existing `a` until the replacement is complete.
- **Preserve non-streaming behavior.** `preprocess()` defaults to the original full trimming; only the streaming parser enables newline preservation. For example, non-streaming input ending with the unfinished string `root = Title("hello` followed by a newline still produces text `"hello"` with `incomplete: true`.
- Add regression coverage and a patch changeset for `@openuidev/lang-core`.

The rendered output now agrees for the duplicate-ID examples. The existing metadata difference remains: `statementCount` counts unique IDs in batch parsing but all definitions in streaming parsing (2 versus 3 in the example above).

## Test Plan

- [x] Parser test file: 37 tests passed, including single-push, line-by-line, character-by-character, two-chunk, `set()`, trailing-newline, and incomplete-replacement cases.
- [x] Direct runtime assertions confirmed that non-streaming unfinished-string newline handling is unchanged and streaming replacements work with and without a final newline.
- [x] ESLint and Prettier checks passed for the modified parser file.

## Checklist

- [x] Linked the related issue.
- [x] Considered backwards compatibility and preserved non-streaming preprocessing.
- [x] Added a patch changeset.

### DIFF
--- a/packages/lang-core/src/parser/__tests__/parser.test.ts
+++ b/packages/lang-core/src/parser/__tests__/parser.test.ts
@@ -287,3 +287,86 @@ root = Title("hello")
     expect(result.root?.props.text).toBe("hello");
   });
 });
+
+// ── redefined statement IDs (parse vs stream) ───────────────────────────────
+
+describe("redefined statement IDs", () => {
+  const withNewline = `root = Stack([a])
+a = Title("x")
+a = Title("y")
+`;
+  const withoutNewline = `root = Stack([a])
+a = Title("x")
+a = Title("y")`;
+
+  const titleText = (result: ReturnType<typeof parse>) => {
+    const children = result.root?.props?.children as any[] | undefined;
+    return children?.[0]?.props?.text as string | undefined;
+  };
+
+  const pushAll = (
+    text: string,
+    write: (sp: ReturnType<typeof createStreamParser>, text: string) => void,
+  ) => {
+    const sp = createStreamParser(schema);
+    write(sp, text);
+    return titleText(sp.getResult());
+  };
+
+  it("parse() last-wins with and without a trailing newline", () => {
+    expect(titleText(parse(withNewline, schema))).toBe("y");
+    expect(titleText(parse(withoutNewline, schema))).toBe("y");
+  });
+
+  it("stream parser last-wins for a single push, matching parse()", () => {
+    expect(pushAll(withNewline, (sp, text) => sp.push(text))).toBe("y");
+    expect(pushAll(withoutNewline, (sp, text) => sp.push(text))).toBe("y");
+  });
+
+  it("stream parser last-wins line-by-line and character-by-character", () => {
+    expect(
+      pushAll(withNewline, (sp, text) => {
+        for (const line of text.split(/(?<=\n)/)) {
+          if (line) sp.push(line);
+        }
+      }),
+    ).toBe("y");
+    expect(
+      pushAll(withNewline, (sp, text) => {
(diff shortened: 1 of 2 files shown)


abhithesys (commented):
the main issue to fix:
1. why only last statement is not replacing origin statement
2. streaming doesn't works for replacing statement

PR #1153, opened 2026-09-10
fix(lang-core): update pristine declaration defaults during streaming (#767)

### LINKED ISSUE #767: bug(lang-core): streaming $binding defaults stick at truncated value (store.initialize skips existing keys)
## Description

In `@openuidev/lang-core`, `createStore().initialize()` only applies declaration defaults for **new** keys. During streaming parse recovery, a `$binding` can first materialize with a **truncated** string default (lexer auto-closes an incomplete `"..."`). That truncated value is written into the store and then **never updated** when the full declaration arrives, because the key already exists.

After a full page reload the store is fresh, so the complete default is applied once and the UI looks correct — which makes the bug easy to miss.

This affects any consumer of the store (`@openuidev/vue-lang`, `@openuidev/react-lang`) that re-calls `initialize` as `stateDeclarations` change during streaming.

## Package / API

- Package: `@openuidev/lang-core`
- API: `createStore` → `initialize(defaults, persisted)` in `packages/lang-core/src/runtime/store.ts`
- Observed via: openui-lang `$bindings` used in `TextContent` / `Mutation` args (e.g. `$title = "..."`)

## Steps to reproduce

1. Render a `<Renderer>` (Vue or React) with streaming `isStreaming` / incremental response text.
2. Stream openui-lang that declares a long string binding, for example:

```text
$title = "issue with OpenUI, missing Vue 3 headless package"
root = Stack([TextContent("Support ticket prepared: " + $title)])
```

3. Observe the first paint while the string literal is still incomplete (lexer recovers by closing the quote early), e.g. `$title` becomes `"issue"`.
4. Let the stream finish so `stateDeclarations.$title` is the full string.
5. Note that the UI still shows the truncated title until a full remount / page reload.

## Expected behavior

Once streaming recovers the complete declaration default for a `$binding` that the user has **not** edited, the store should update to the new default so the UI shows the full value without reload.

## Actual behavior

- Mid-stream: `$title === "issue"` (or another truncated recovery).
- After stream completes: `$title` stays truncated because `initialize` skips existing keys.
- After page reload: `$title` is correct (fresh store).

Relevant current logic in [`packages/lang-core/src/runtime/store.ts` L84–L88](https://github.com/thesysdev/openui/blob/9d3be653719bb91ef2988f49b0eff9ce9bc74f30/packages/lang-core/src/runtime/store.ts#L84-L88):

```ts
for (const key of Object.keys(defaults)) {
  if (!state.has(key)) {
    state.set(key, defaults[key]);
  }
}
```

The comment intends to preserve user-edited bindings and avoid deleting keys when declarations briefly disappear during streaming — but it also blocks updating **pristine** declaration defaults as they grow from partial → complete.

## Proposed direction (for maintainer approval)

Track pristine keys (set only via `initialize`, cleared on `set()` / persisted restore). On re-`initialize`, update values for keys that are still pristine when the declaration default changes; keep dirty / user-edited values untouched.

Happy to open a PR once this issue is approved (per CONTRIBUTING).

## Environment

- Monorepo: `thesysdev/openui`
- Package: `@openuidev/lang-core` (workspace)
- Reproduced in `examples/vue-chat` with Query/Mutation + `$title` bindings while streaming assistant messages
- Node / pnpm workspace setup as in repo docs

## Additional context

- Relates to streaming lexer recovery that closes incomplete string tokens (`packages/lang-core/src/parser/lexer.ts`).
- Same `initialize` is used from both `vue-lang` and `react-lang` state hooks when `stateDeclarations` change.

justonemorenight:
### Summary

Fixes #767

In `@openuidev/lang-core`'s reactive state store (`createStore().initialize()`), declaration defaults previously only applied to keys not yet present in `state` (`!state.has(key)`).

During streaming parse recovery, a `$binding` (e.g. `$title = "..."`) can first materialize with a truncated string default when the lexer auto-closes an incomplete open quote. When the stream progresses and the full declaration arrives, `initialize` skipped updating the key because it already existed in `state`, leaving the UI stuck at the truncated value until a full reload.

### Changes

- In `packages/lang-core/src/runtime/store.ts`:
  - Track pristine keys (keys seeded from `defaults` that have not been modified by `store.set()` or initialized via `persisted`).
  - When `initialize(defaults, persisted)` runs:
    - Persisted entries remove keys from `pristine` and update `state`.
    - New keys are added to `state` and marked `pristine`.
    - Existing keys that are still in `pristine` (unmodified by the user) update to the latest declaration default if it changed.
    - User-modified values (`set()` removes from `pristine`) and persisted values remain untouched.
  - `store.dispose()` clears the `pristine` tracking set.
- Added comprehensive unit tests in `packages/lang-core/src/runtime/__tests__/store.test.ts` verifying:
  - Pristine defaults update as streaming declarations grow.
  - User edits via `store.set()` are never overwritten by subsequent default updates.
  - Persisted values are never overwritten by subsequent default updates.
  - Listeners are notified only when values actually change.
  - Dispose clears state and pristine sets cleanly.

### Testing

- `pnpm --filter @openuidev/lang-core test` — all 8 test files (118 tests) passed.
- `pnpm --filter @openuidev/react-lang test` — all 3 test files (12 tests) passed.
- `pnpm --filter @openuidev/vue-lang test` — all 4 test files (36 tests) passed.
- Lint and Prettier checks passed cleanly.

### DIFF
--- a/packages/lang-core/src/runtime/store.ts
+++ b/packages/lang-core/src/runtime/store.ts
@@ -13,6 +13,7 @@ export interface Store {
 
 export function createStore(): Store {
   const state = new Map<string, unknown>();
+  const pristine = new Set<string>();
   const listeners = new Set<() => void>();
   let snapshot: Record<string, unknown> = {};
 
@@ -57,6 +58,7 @@ export function createStore(): Store {
         return;
       }
     }
+    pristine.delete(name);
     state.set(name, value);
     rebuildSnapshot();
     notify();
@@ -74,24 +76,46 @@ export function createStore(): Store {
   }
 
   function initialize(defaults: Record<string, unknown>, persisted: Record<string, unknown>): void {
-    // Apply persisted values (explicit restore) and defaults for NEW keys only.
-    // Existing user-modified $binding values are always preserved — never
-    // overwrite with defaults, never delete. During streaming, declarations
-    // can temporarily disappear; deleting user state here would cause data loss.
+    let changed = false;
+
+    // Apply persisted values (explicit restore). These take precedence and are not pristine.
     for (const key of Object.keys(persisted)) {
-      state.set(key, persisted[key]);
-    }
-    for (const key of Object.keys(defaults)) {
-      if (!state.has(key)) {
-        state.set(key, defaults[key]);
+      pristine.delete(key);
+      const val = persisted[key];
+      if (!state.has(key) || !Object.is(state.get(key), val)) {
+        state.set(key, val);
+        changed = true;
       }
     }
-    rebuildSnapshot();
-    notify();
+
+    // Apply defaults:
+    // 1. For newly encountered keys, seed default and mark as pristine.
+    // 2. For existing keys that are STILL pristine (never modified by user or persisted),
+    //    allow updated declaration defaults to take effect (e.g. streaming string recovery).
+    // Existing user-modified values (not in pristine) are always preserved.
+    for (const key of Object.keys(defaults)) {
+      const val = defaults[key];
+      if (!state.has(key)) {
+        state.set(key, val);
+        pristine.add(key);
+        changed = true;
+      } else if (pristine.has(key)) {
+        if (!Object.is(state.get(key), val)) {
+          state.set(key, val);
+          changed = true;
+        }
+      }
+    }
+
+    if (changed) {
+      rebuildSnapshot();
+      notify();
+    }
   }
 
   function dispose(): void {
     state.clear();
+    pristine.clear();
     listeners.clear();
     snapshot = {};
   }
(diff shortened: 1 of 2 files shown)


packages/lang-core/src/runtime/store.ts
```diff
@@ -74,24 +76,46 @@ export function createStore(): Store {
...
+    // Existing user-modified values (not in pristine) are always preserved.
     for (const key of Object.keys(defaults)) {
+      const val = defaults[key];
       if (!state.has(key)) {
-        state.set(key, defaults[key]);
+        state.set(key, val);
+        pristine.add(key);
+        changed = true;
+      } else if (pristine.has(key)) {
+        if (!Object.is(state.get(key), val)) {
```
Aditya-thesys: can you please use the same shallow comparision as the above `set` function

Aditya-thesys:
Thanks! for the fix, all looks good. Just fix the small nit and it is good to go.

PR #1150, opened 2026-09-10
docs: add the Auto-fix API to the Gateway docs

ankur-mittal95:
## Why

The Gateway is getting a standalone endpoint that repairs invalid OpenUI Lang without routing the request through generation (thesysdev/muse#599, thesysdev/coda#1024). Users who call a model directly need to find it from the Gateway docs and know what it takes, returns, and costs.

The endpoint ships with thesysdev/muse#599 and thesysdev/coda#1024. Merge this once those are deployed, so the page never describes something that does not answer.

## What changes

- **New page, Auto-fix API** (`gateway/api/auto-fix.mdx`): endpoint, a complete example with the library spec import and the API key, the request fields, the three `status` values with a sample response for a fix and for a failed fix, the error codes with their meaning, limits, a pricing pointer, HTTP errors, and how calls appear in the console.
- **Introduction**: a third card under "Choose an API" and one bullet under "What OpenUI Gateway provides".
- **Sidebar**: `Auto-fix API` under APIs, after Conversations API.
- **Reliability**: a "New: Auto-fix API" callout at the top of the page, the table row updated to mention it, and one line under "What your application still controls" about fallback when a repair fails.
- **Pricing**: a short note under Gateway rates. Flat $0.02 per call that runs a fix, from Gateway Credits; a valid generation is free; no plan API call. The API page only links here.

Wording follows the existing pages: short sentences, the reader's side of the screen, and the same section order as the other API pages. The existing pages were edited in place without reformatting, since they are not prettier-formatted.

## Check before merging

- The sample responses were captured from a local run and match the contract in the muse PR.
- Rendered locally on the docs dev server: intro cards, sidebar entry, the new page, the Reliability callout, and the Pricing note all display as intended.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Review round 1

Applied: no "program" anywhere (the API field is now `generation` on both sides); `openUIlibraryVersion` and `customActions` removed, `library` is the one input and takes `openui.spec.json` as the CLI writes it; full examples with imports; artifact limit line removed; "two attempts to fix the generation"; price stated only on the Pricing page.

## Review round 2

Renamed the API from Sanitize to Auto-fix: page `api/auto-fix.mdx`, sidebar entry, card, callout, endpoint `/v1/embed/auto-fix`, console model name `openui/auto-fix`.

## Conversation context

[...]

PR #1162, opened 2026-09-11
docs: the Auto-fix API takes the conversation

ankur-mittal95:
Stacked on #1150. Review that one first.

The Auto-fix API can now take the conversation that produced the generation
(thesysdev/muse#601). The repair reads it to understand what the user asked
for, so a truncated list keeps the rows the user wanted and an unknown
component is replaced by one that still answers the question.

Changes on the Auto-fix page:

- `messages` is in the Request table, optional, an array of `{ role, content }`.
- The `lib/auto-fix.ts` and `server.ts` examples pass the same messages the
  caller sent to their model.
- Limits say 20 turns, 8,000 characters in total, and that older turns are
  dropped first.
- A short "Data" note says the conversation goes to the model that fixes the
  generation and is not stored beyond the request logs.
- The opening paragraph no longer says the request has no conversation, and the
  400 row names a conversation over a limit.

No other file is touched, and the prose style is unchanged.

### DIFF
--- a/docs/content/docs/gateway/api/auto-fix.mdx
+++ b/docs/content/docs/gateway/api/auto-fix.mdx
@@ -5,7 +5,7 @@ description: Fix invalid OpenUI Lang from any model with one API call.
 
 The Auto-fix API fixes OpenUI Lang after a model has generated it. Send the raw generation and your component library. The API validates the generation, fixes the errors it can, and returns valid OpenUI Lang that the renderer can use.
 
-Use it when your application calls a model directly and only needs the correction step. The request does not name a model, and nothing new is generated. The API applies the same correction that runs inside the [Chat Completions](/docs/gateway/api/chat-completions) and [Responses](/docs/gateway/api/responses) APIs, as a standalone call.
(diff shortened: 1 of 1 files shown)


docs/content/docs/gateway/api/auto-fix.mdx
```diff
@@ -5,7 +5,7 @@ description: Fix invalid OpenUI Lang from any model without routing the request
 
 The Auto-fix API fixes OpenUI Lang after a model has generated it. Send the raw generation and your component library, and Gateway validates the generation, fixes the errors it can, and returns valid OpenUI Lang that the renderer can use.
 
-Use it when your application calls a model directly and only needs the correction step. The request does not include a conversation or a model, and Gateway does not generate anything new. It applies the same correction that runs inside the [Chat Completions](/docs/gateway/api/chat-completions) and [Responses](/docs/gateway/api/responses) APIs, as a standalone call.
+Use it when your application calls a model directly and only needs the correction step. The request does not name a model, and Gateway does not generate anything new. It applies the same correction that runs inside the [Chat Completions](/docs/gateway/api/chat-completions) and [Responses](/docs/gateway/api/responses) APIs, as a standalone call. You can also send the conversation that led to the generation, and Gateway reads it only to understand what the user asked for.
```
ankur-mittal95: let's not mention gateway

